PR #16, opened 2022-01-21
Remove duplicate lifecycle-reactivestreams-ktx dependency in samples.

valdesekamdem:
## Description
Remove duplicate lifecycle-reactivestreams-ktx dependency in viewmodels-declarative sample for Android.

## Motivation and Context
I noticed a duplicate dependency in trickot-viewmodels-declarative sample for Android, seems to be a copy/paste error.


marcantoinefortier (approved):
Good catch! I was getting a warning too in my project 👍🏻

PR #19, opened 2022-01-28
[Experimental] Migrate to Kotlin 1.6 new Memory Model

marcantoinefortier:
⚠️ IMPORTANT! Before migrating to the new memory model in your project, please read the whole [official Kotlin documentation](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md#new-memory-model-migration-guide) to understand all the implications following this migration.
> The new memory model is Experimental. It's not production-ready and may be changed at any time. Opt-in is required (see the details below), and you should use it only for evaluation purposes.

📌 The obective of this PR is for other teams to be able to migrate to this new memory model and test the impact it has on their  project.

## Description

Following the new memory model migration guide https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md :

- Remove all freezing 🧊
- Update to [Ktor 2.0.0-beta-1](https://github.com/ktorio/ktor/releases/tag/2.0.0-beta-1) 🚀
- Use [kotlinx.atomicfu](https://github.com/Kotlin/kotlinx.atomicfu) for Native AtomicReference implementation ⚛️

## Motivation and Context

Freezing has a major cost on performance (especially on the ui dispatch queue aka the main thread) as it needs to freeze the whole subgraph of the object. We hope the garbage collection will reduce the effects of freezing subgraphs and offer a more "stable" memory management.

## How Has This Been Tested?

This has been tested using the latest version of a production app built here at [Mirego](github.com/mirego). Again, the goal is to test this migration on as many projects as possible to better assert the implications of this change.

## Types of changes


Trikot.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
   spec.name          = "Trikot"
-  spec.version       = "1.0.0"
+  spec.version       = "3.2.0-M1"
```
npresseault, line 3: What's that?

trikot-foundation/trikotFoundation/src/nativeMain/kotlin/com/mirego/trikot/foundation/concurrent/AtomicReference.kt
```diff
@@ -1,45 +1,13 @@
 package com.mirego.trikot.foundation.concurrent
 
-import kotlin.native.concurrent.freeze
-import kotlin.native.concurrent.isFrozen
-
 actual class AtomicReference<T> actual constructor(value: T) {
```
npresseault, line 3: This should be the baseline implementation for all targets. And eventually we can deprecate it in favor of using atomic-fo directly.

npresseault (approved):
Would be nice to have something in the changelog for breaking changes (aka MrFreeze gone).

Trikot.podspec
marcantoinefortier: I was triggered when installing pods, where the Podspec version would not expose the actual version of the lib 😅 : 
```
Installing Trikot (1.0.0)
```

I feel like we should instead see :
```
Installing Trikot (3.2.0-M1)
```

npresseault: maybe we could leverage the podspec_versions script and use the properties file version as a single source of truth?

npresseault: I was referring to this one
```
def load_properties(properties_filename)
  properties = {}
  File.open(properties_filename, 'r') do |properties_file|
    properties_file.read.each_line do |line|
      line.strip!
      if (line[0] != ?# and line[0] != ?=)
        i = line.index('=')
        if (i)
          properties[line[0..i - 1].strip] = line[i + 1..-1].strip
        else
          properties[line] = ''
        end
      end
    end
  end
  properties
end

```

marcantoinefortier: Actually we should migrate to the [Gradle Cocoapods Plugin](https://plugins.gradle.org/plugin/org.jetbrains.kotlin.native.cocoapods) and the versionning will be free! 😄

trikot-bluetooth/swift-extensions/TrikotAttributeProfileService.swift
```diff
@@ -7,7 +7,7 @@ class TrikotAttributeProfileService: NSObject, AttributeProfileService {
 
     var trikotCharacteristics = [CBCharacteristic: TrikotAttributeProfileCharacteristic]()
 
-    let characteristics: Publisher = frozenBehaviorSubject()
+    let characteristics: Publisher = Publishers.behaviorSubject(value: nil)
```
jpmartin, line 10:
```suggestion
    let characteristics: Publisher = Publishers().behaviorSubject(value: nil)
```

trikot-bluetooth/swift-extensions/TrikotBluetoothDevice.swift
```diff
@@ -10,8 +10,8 @@ class TrikotBluetoothDevice: NSObject, BluetoothDevice, CBPeripheralDelegate {
 
     var physicalAddress: String
 
-    var attributeProfileServices: Publisher = frozenBehaviorSubject()
-    var isConnected: Publisher = frozenBehaviorSubject()
+    var attributeProfileServices: Publisher = Publishers.behaviorSubject(value: nil)
+    var isConnected: Publisher = Publishers.behaviorSubject(value: nil)
```
jpmartin, line 14:
```suggestion
    var attributeProfileServices: Publisher = Publishers().behaviorSubject(value: nil)
    var isConnected: Publisher = Publishers().behaviorSubject(value: nil)
```

trikot-bluetooth/swift-extensions/TrikotBluetoothManager.swift
```diff
@@ -7,14 +7,14 @@ public class TrikotBluetoothManager: NSObject, BluetoothManager, CBCentralManage
     private let dispatchQueue = DispatchQueue(label: TrikotBluetoothManager.QueueLabel, qos: .background)
     private lazy var centralManager = CBCentralManager(delegate: self, queue: dispatchQueue)
 
-    private let scanResultpublisher = frozenBehaviorSubject()
+    private let scanResultpublisher = Publishers.behaviorSubject(value: nil)
```
jpmartin, line 10:
```suggestion
    private let scanResultpublisher = Publishers().behaviorSubject(value: nil)
```

trikot-bluetooth/swift-extensions/TrikotBluetoothManager.swift
```diff
@@ -7,14 +7,14 @@ public class TrikotBluetoothManager: NSObject, BluetoothManager, CBCentralManage
...
     private let numberOfScanRequest = Atomic<Int>(0)
 
     private var scanResults = [CBPeripheral: TrikotBluetoothScanResult]()
     private var connectedDevices = [CBPeripheral: TrikotBluetoothDevice]()
     private var delegates = [CBCentralManagerDelegate]()
 
-    private let bluetoothStatePublisher: Publisher = frozenBehaviorSubject()
+    private let bluetoothStatePublisher: Publisher = Publishers.behaviorSubject(value: nil)
```
jpmartin, line 17:
```suggestion
    private let bluetoothStatePublisher: Publisher = Publishers().behaviorSubject(value: nil)
```

trikot-bluetooth/swift-extensions/TrikotBluetoothScanResult.swift
```diff
@@ -6,15 +6,15 @@ class TrikotBluetoothScanResult: NSObject, BluetoothScanResult {
     private let centralManager: CBCentralManager
     private let peripheral: CBPeripheral
     private var lostHeartbeatTimer: Foundation.Timer?
-    private var manufacturerData = frozenBehaviorSubject()
+    private var manufacturerData = Publishers.behaviorSubject(value: nil)
```
jpmartin, line 9:
```suggestion
    private var manufacturerData = Publishers().behaviorSubject(value: nil)
```

trikot-bluetooth/swift-extensions/TrikotBluetoothScanResult.swift
```diff
@@ -6,15 +6,15 @@ class TrikotBluetoothScanResult: NSObject, BluetoothScanResult {
...
 
     var bluetoothDevice: TrikotBluetoothDevice?
 
     var name: String
 
     var physicalAddress: String
 
-    var rssi: Publisher = frozenBehaviorSubject()
+    var rssi: Publisher = Publishers.behaviorSubject(value: nil)
```
jpmartin, line 17:
```suggestion
    var rssi: Publisher = Publishers().behaviorSubject(value: nil)
```

trikot-bluetooth/swift-extensions/TrikotAttributeProfileCharacteristic.swift
```diff
@@ -49,4 +49,3 @@ class TrikotAttributeProfileCharacteristic: NSObject, AttributeProfileCharacteri
         super.init()        
     }
 }
-
```
jpmartin, line 52: At line 7, there is a `let event = frozenSubject()` that needs to be replaced with: `let event: Publisher = Publishers().publishSubject()`

jpmartin (approved):
Good job! 👏

martingagnon (approved):
💯 👏

PR #21, opened 2022-02-02
Add accessibility properties to ViewModel

CharlesMcD:
## Description

It was not possible to manage accessibility with our viewModel. While doing it in our project I took the opportunity to bring it in trikot.viewModel. 

Bonus : I also fixed the sample app that had a cell reuse problem on iOS.

## Motivation and Context

- `isAccessibilityElement: Publisher<Boolean>` : Return true if it should be exposed as an accessibility element.
- `accessibilityLabel: Publisher<String>` : String that identifies the accessibility element.
- `accessibilityHint: Publisher<ViewModelAccessibilityHint>` : Describes the result of performing an action on the element.

## How Has This Been Tested?

In the sample app

**Android**

https://user-images.githubusercontent.com/6673075/152240065-bfa50493-1563-4805-aef9-b352c4c0b64a.mp4

**iOS**

https://user-images.githubusercontent.com/6673075/152240098-8d47557c-699a-418c-93f8-a69801701494.mp4

## Types of changes


trikot-viewmodels/viewmodels/src/commonMain/kotlin/com/mirego/trikot/viewmodels/ViewModel.kt
```diff
@@ -23,4 +23,18 @@ interface ViewModel {
      * Action to execute when the view is tapped
      */
     val action: Publisher<ViewModelAction>
+    /**
+     * Return true if it should be exposed as an accessibility element.
+     * Note : `accessibilityLabel` or `accessibilityHint` must be specified (at least one) if this property is set to true or it will have no impact.
+     */
+    val isAccessibilityElement: Publisher<Boolean>
```
martingagnon, line 30: I think `isAccessibilityElement` could be a get var where we combine both label and hint and return true if one emit.

```
val isAccessibilityElement:  get { accessibilityHint.combine(accessibilityLabel).map { true }.first().shared() .startWith(false) }
```

martingagnon: As it is a rule (that at least one emit something) I would enforce this this way.

trikot-viewmodels/viewmodels/src/commonMain/kotlin/com/mirego/trikot/viewmodels/ViewModel.kt
```diff
@@ -23,4 +23,18 @@ interface ViewModel {
...
+    /**
+     * String that identifies the accessibility element
+     */
+    val accessibilityLabel: Publisher<String>
+    /**
+     * Describes the result of performing an action on the element, when the result is non-obvious.
+     * For example: "Purchases the item." or "Downloads the attachment."
+     */
+    val accessibilityHint: Publisher<String>
```
marcantoinefortier, line 39: These should all be part of another interface, not the on ViewModel itself. Let's say :

```kotlin
interface Accessible {
    val accessibilityLabel: Publisher<String>
    val accessibilityHint: Publisher<String>
```

marcantoinefortier: Also, I feel like these should not be publishers, but instead static elements. It seems weird for the user to have inconsistent values when running through accessibility elements. What do you think @dominiclabbe ?

trikot-viewmodels/viewmodels/src/commonMain/kotlin/com/mirego/trikot/viewmodels/ViewModel.kt
CharlesMcD: Having used it in "real life", it happens quite a lot that we want to temporarily make an element inaccessible. An example is when we display a side panel and we want this panel to be the only accessible element, we can disable everything else. 
IMO it seems easier and straightforward to simply emit a new boolean instead of reinitializing the `accessibilityLabel` and the `accessibilityHint` every time we want to make something inaccessible.

trikot-viewmodels/viewmodels/src/commonMain/kotlin/com/mirego/trikot/viewmodels/ViewModel.kt
CharlesMcD: All those values can change frequently. For example, the same button that is used to play and pause a video will have a different `accessibilityHint` each time we interact with it in the same way we alternate the resources between the "pause" and "pause" image.

I agree however that we could extract the new accessibilities properties in an `Accessible` interface like you propose. 👍 Or maybe something like `ViewModelAccessible` to limit the potential name conflicts?

trikot-viewmodels/sample/ios/iosApp/helpers/AutosizingCell.swift
```diff
@@ -1,15 +1,26 @@
...
+class AutosizingCell<T: UIView>: UITableViewCell {
+    var view = T()
 
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
+
+        configureView()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
```
thermech, line 12: `@available(*, unavailable)`

thermech (commented):
👍

trikot-viewmodels/viewmodels/src/commonMain/kotlin/com/mirego/trikot/viewmodels/ViewModel.kt
marcantoinefortier: You're right, it actually makes sense with reactive components. 👍🏻 It might be different with declarative UI, though 👀

marcantoinefortier (approved):

trikot-viewmodels/sample/common/src/commonMain/kotlin/com/trikot/viewmodels/sample/viewmodels/home/ViewsViewModel.kt
```diff
@@ -18,20 +18,42 @@ class ViewsViewModel : MutableListViewModel<ListItemViewModel>() {
...
+        },
+        MutableHeaderListItemViewModel(".accessibilityLabel (\"Sample view\")"),
+        MutableViewListItemViewModel().apply {
+            view.accessibilityLabel = "This is a sample view".just()
+            view.isAccessibilityElement = true.just()
+        },
+        MutableHeaderListItemViewModel(".accessibilityHint (\"Purchase the item.\")"),
+        MutableViewListItemViewModel().apply {
+            view.accessibilityLabel = "Purchase".just()
+            view.accessibilityHint = "Purchase the item".just()
```
marcantoinefortier, line 54: I would add a test using a behaviourSubject that actually changes the accessibility components after purchasing, just to showcase that these can be updated dynamically.

trikot-viewmodels/viewmodels/src/commonMain/kotlin/com/mirego/trikot/viewmodels/ViewModel.kt
dominiclabbe: That's what we did in Fibe, every component we want to make accessible implements the `Accessible` interface.

trikot-viewmodels/sample/common/src/commonMain/kotlin/com/trikot/viewmodels/sample/viewmodels/home/ViewsViewModel.kt
CharlesMcD: Good thing you mentioned this because I realized that `accessibilityHint` lacked flexibility. 
Updating the `accessibilityLabel` worked great but the new `accessibilityHint` was not announced right away when it was modified. (We had to unfocus and focus again to know the new value)

I've made some changes that allow us to specify the new hint, also state if we want the new value to be announced immediately and if we want a `customHintsChangeAnnouncement` just for this moment.

So for example the normal `hint` can be "Purchase the item" / "Cancel your purchase" and the `customHintsChangeAnnouncement` can be "Interact again to cancel your purchase".

https://user-images.githubusercontent.com/6673075/152411967-633d5a36-20b9-43c0-ac8b-2c4f9cd2127e.mp4

https://user-images.githubusercontent.com/6673075/152411996-4d3ecbe5-0a5b-4781-b89a-5638dc3eabe5.mp4

trikot-viewmodels/viewmodels/src/androidMain/kotlin/com/mirego/trikot/viewmodels/ViewModelBinder.kt
```diff
@@ -35,6 +36,18 @@ fun View.bindViewModel(
             setAlpha(alpha)
         }
 
+        it.isAccessibilityElement.observe(lifecycleOwnerWrapper.lifecycleOwner) { isAccessibilityElement ->
+            importantForAccessibility = if (isAccessibilityElement) View.IMPORTANT_FOR_ACCESSIBILITY_YES else View.IMPORTANT_FOR_ACCESSIBILITY_NO
+        }
```
dominiclabbe, line 41: I'm not sure how this will behave with a custom view, that contains sub views. For instance, we don't want a custom button composed of a Frame an icon and a label to have 3 accessibility nodes that can be focused by the screen reader. In a case like this, only the parent Frame view should be important for accessibility, and have a contentDescription that describes the button and its state.

dominiclabbe: If we don't set isAccessibilityElement when initializing the ViewModel, is this code called? If not, it would leave the default value for importantForAccessibility and that would be ok.

CharlesMcD: We could do something like this right? The `MyCustomView` will be accessible but not the subviews.

```
class MyCustomView {
    myLabel: LabelViewModel
    myImage: ImageViewModel

    init() {
        isAccessibilityElement = true
        myLabel.isAccessibilityElement = false
        myImage.isAccessibilityElement = false
    }
}
```

CharlesMcD: > If we don't set isAccessibilityElement when initializing the ViewModel, is this code called?

@dominiclabbe no it's not called since the default value in `MutableViewModel` is 
```
override var isAccessibilityElement = PropertyFactory.never<Boolean>()

override var accessibilityLabel = PropertyFactory.never<String>()

override var accessibilityHint = PropertyFactory.never<ViewModelAccessibilityHint>()
```

dominiclabbe: Ok, I think that could do it then.

marcantoinefortier (approved):
Good job @CharlesMcD 💯 

I'm glad we're going towards better accessibility in our apps, this will make it a lot easier to do when it's already built in 🧠

dominiclabbe:
From the videos, we don't see any of the buttons being described as "button", because the view model is bound to plain views. One way to fix this is to force the accessibility role or trait.

On iOS, you simply add the `UIAccessibilityTraitButton` trait to the accessibilityTraits value:
`view.accessibilityTraits |= .button`

On Android, it's a bit more complicated, you have to use an `accessibilityDelegate` to force the class name, or override `view.onInitializeAccessibilityNodeInfo`
```
ViewCompat.setAccessibilityDelegate(model, object : AccessibilityDelegateCompat() {
    override fun onInitializeAccessibilityNodeInfo(host: View, info: AccessibilityNodeInfoCompat) {
        super.onInitializeAccessibilityNodeInfo(host, info)
        info.className = android.widget.Button::class.java.name
    }
```

This wouldn't be necessary it the button view model were attached to actual button views.

CharlesMcD:
> From the videos, we don't see any of the buttons being described as "button", because the view model is bound to plain views. One way to fix this is to force the accessibility role or trait.
> 
> On iOS, you simply add the `UIAccessibilityTraitButton` trait to the accessibilityTraits value: `view.accessibilityTraits |= .button`
> 
> On Android, it's a bit more complicated, you have to use an `accessibilityDelegate` to force the class name, or override `view.onInitializeAccessibilityNodeInfo`
> 
> ```
> ViewCompat.setAccessibilityDelegate(model, object : AccessibilityDelegateCompat() {
>     override fun onInitializeAccessibilityNodeInfo(host: View, info: AccessibilityNodeInfoCompat) {
>         super.onInitializeAccessibilityNodeInfo(host, info)
>         info.className = android.widget.Button::class.java.name
>     }
> ```
> 
> This wouldn't be necessary it the button view model were attached to actual button views.

Good point! Since this is an addition and not a modification I will extract your comment in a github issue/task so that we can add it eventually. There is also the ([accessibilityValue](https://developer.apple.com/documentation/uikit/uiaccessibilityelement/1619583-accessibilityvalue)) that would be interesting to consider.

dominiclabbe (commented):
Great start! 👌🏼

antoinelamy:
Turns out that `trikot.viewmodels` has a deployment target of iOS/tvOS 9 and that `.accessibilitySpeechQueueAnnouncement` is iOS/tvOS 11+, failing the build unless we override the pod deployment target in the Podfile.

PR #27, opened 2022-02-11
Fixed the method to observe a ViewModel as State (Jetpack Compose)

gbourassa:
When observing a ViewModel asState, we now use the instance of the ViewModel as the key to decide if remember() must re-calculate the value.

## Motivation and Context

The current implementation of observeAsState() uses the remember mechanism to remember the value.  The mechanism behind Remember() is that the value is "calculated" once, and if the UI is re-composed afterward, the same value will be used.

The problem with that approach here is that if we observeAsState a ViewModel, and the _instance_ of the viewmodel changes between two recomposition, the remembered state will not be recalculated with the new instance.

By using the ViewModel instance as the "rememberance" key, we make sure that we update the State.

## Types of changes


js-bonin (approved):
💯

antoinelamy (approved):
👌

PR #31, opened 2022-02-16
[VMD] Add the possibility to animate object change in ObservableViewModelAdapter

hugolefrancois:
## Description

This is a small first step to be able to animate modification triggered by the common code. By passing an animation to the ObservableViewModelAdapter, all changes of the object will be animated.

## How Has This Been Tested?

In my project

## Types of changes

### DIFF
--- a/trikot-viewmodels-declarative/swift/swiftui/Utilities/ObservableViewModelAdapter.swift
+++ b/trikot-viewmodels-declarative/swift/swiftui/Utilities/ObservableViewModelAdapter.swift
@@ -1,4 +1,4 @@
-import UIKit
+import SwiftUI
 import Combine
 import TRIKOT_FRAMEWORK_NAME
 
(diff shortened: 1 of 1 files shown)


OlivierPineau (approved):
Very nice

martingagnon (approved):
💯

trikot-viewmodels-declarative/swift/swiftui/Utilities/ObservableViewModelAdapter.swift
```diff
@@ -7,13 +7,19 @@ public class ObservableViewModelAdapter<VM: VMDViewModel>: ObservableObject {
     private let propertyWillChangePublisher: AnyPublisher<KotlinUnit, Never>
     private var cancellable: AnyCancellable?
 
-    public init(viewModel: VM) {
+    public init(viewModel: VM, animation: Animation? = nil) {
```
antoinelamy, line 10: I don't think that globally set animations for any property change is a good idea. We should be able to decide, for a given property change if we want that change to be animated or not. Animating everything would cause more harm than good IMO.

marcantoinefortier (approved):
An example would have been a nice addition in the sample app 🙂

PR #32, opened 2022-02-17
[VMD] Fix android RemoteImage to allow alignment + add a placeholder content scale property

js-bonin:
## Description
There was an extra Box layer wrapping the RemoteImage which was not being passed the received modifier so when passing a modifier with align properties the image was align within that extra box layer instead of within the VMDImage parent's layout.

We also add an option to add a different content scale for the placeholder image.

## Motivation and Context
 With the current implementation it is not possible to correctly align a VMDImage that is of RemoteImage type.

## How Has This Been Tested?
Tested with the sample app.

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)


martingagnon (approved):
LGTM!

antoinelamy (approved):
Great addition 👍🏻

PR #42, opened 2022-03-04
[VMD] Fix crash in VMDButtonViewModel

hugolefrancois:
## Description

Having the VMDButtonViewModel actionBlock as a `var` was causing freezing exception if the ViewModel was freeze before setting the action. The ViewModel was frozen more often than we imagined, only binding a property on the ViewModel is freezing the ViewModel and all of the objects around 😬

## Motivation and Context

A regression was introduce in a previous modification and that regression was generating crashes everywhere 😬

## How Has This Been Tested?

Tested in VMD sample project

## Types of changes

### DIFF
--- a/trikot-viewmodels-declarative/viewmodels-declarative/src/commonMain/kotlin/com/mirego/trikot/viewmodels/declarative/components/impl/VMDButtonViewModelImpl.kt
+++ b/trikot-viewmodels-declarative/viewmodels-declarative/src/commonMain/kotlin/com/mirego/trikot/viewmodels/declarative/components/impl/VMDButtonViewModelImpl.kt
@@ -1,5 +1,6 @@
 package com.mirego.trikot.viewmodels.declarative.components.impl
 
+import com.mirego.trikot.foundation.concurrent.atomic
 import com.mirego.trikot.streams.cancellable.CancellableManager
 import com.mirego.trikot.streams.reactive.first
 import com.mirego.trikot.streams.reactive.observeOn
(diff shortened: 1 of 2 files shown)


trikot-viewmodels-declarative/viewmodels-declarative/src/commonMain/kotlin/com/mirego/trikot/viewmodels/declarative/components/impl/VMDButtonViewModelImpl.kt
```diff
@@ -17,19 +19,19 @@ open class VMDButtonViewModelImpl<C : VMDContent>(
     defaultContent: C
 ) : VMDButtonViewModel<C>(cancellableManager) {
 
-    override var actionBlock: () -> Unit = {}
+    private val atomicReferenceActionBlock = AtomicReference { }
+
+    override val actionBlock: () -> Unit get() = atomicReferenceActionBlock.value
```
marcantoinefortier, line 24:
```suggestion
    override var actionBlock: () -> Unit by atomic { }
```

You could use the `by atomic()` extension to keep using a var without the need of a separate variable

marcantoinefortier (approved):
You could also update to [Trikot v3.2.0-M1](https://github.com/mirego/trikot/releases/tag/3.2.0-M1) if you'd like to be able to use vars in common code without the need to freeze it 😄

antoinelamy (approved):

antoinelamy:
> You could also update to [Trikot v3.2.0-M1](https://github.com/mirego/trikot/releases/tag/3.2.0-M1) if you'd like to be able to use vars in common code without the need to freeze it 😄

@marcantoinefortier we cannot "update" trikot because it is trikot 😅. The fix could be undone in the M1 branch following the rebase though.

marcantoinefortier:
@antoinelamy Yeah I meant you could update *your project* to Trikot 3.2.0-M1 to avoid the need of freezing 😅

PR #47, opened 2022-03-14
[VMD] Animation engine

antoinelamy:
## Description
This PR provides an engine to drive animations from common code.

It supports all the basic types of animations:
- Ease In
- Ease Out
- Ease In/Out
- Custom Cubic Bezier
- Spring

This PR also:
- Refactor sample's Home Screen to include sections
- Adds a SwiftUI utility to use `ForEach` directly with `VMDIdentifiableContent`
- Adds `AtomicStackReference` in `trikot.foundation`

## Motivation and Context
Using the MVVM pattern it is pretty hard for a view to perform a change animated because the view layer don't have any insight on the type and context of a change. To alleviate that, we introduce the concept of animations in the view model layer.

There is basically two ways to perform an animated change:

### Closure based
A closure based version suitable when the change occur as a result of the press of a button or a specific lifecycle event. All the published properties updated within that block will be animated.
```kotlin
withAnimation(VMDAnimation.Tween(duration = 1.seconds, easing = VMDAnimationEasing.Standard.EaseInEaseOut)) {
        // Perform the change here
}
```

### Binding based
If the change is introduced as part of a binding on a reactive chain, we also have the possibility to bind a value animated. Instead of providing only the value, we have to supply a pair of the value and an optional animation object. The exemple bellow animates the alpha value but does not animate the initial value and animates only the `visible` -> `hidden` transition while performing the `hidden` -> `visible` change without animation.

```kotlin
bindHiddenAnimated(
    hiddenPublisher.withPreviousValue().map {
        val previousValue = it.first
        val value = it.second
        if (previousValue == null) { // No animation if its the first value received
            Pair(value, null)
        } else { // Optionally animate only some changes
            Pair(value, if (value) VMDAnimation.Tween() else null)
        }
    }
)
```

#### SwiftUI
Since SwiftUI is able to automatically infer the updated properties and interpolate between them, there is no further step required for the animation to be effective. It is the equivalent of performing a change for a `@State` or an `@ObservedObject` within a `withAnimation` closure.

#### Jetpack Compose
It is a bit more effort on the Jetpack Compose side because the animation engine requires that we specifically declare an interpolator for a property to be animated.

**Example for animation of the alpha value**
```kotlin
val animatedHiddenProperty by viewModel.textField.observeAnimatedPropertyAsState(
    property = viewModel.textField::isHidden,
    transform = { hidden -> if (hidden) 0f else 1f }
)
val alphaAnimationProgress by animateFloatAsState(
    targetValue = animatedHiddenProperty.value,
    animationSpec = animatedHiddenProperty.animationSpec()
)
````

## Notes
The previous animation mechanism that was iOS only and that would animate all the changes no mater the type of change have been completely removed as it is nearly unused internally and it has just been introduced.

### iOS
https://user-images.githubusercontent.com/291573/158255503-7b985480-776b-4c66-9726-54281c2c8c48.mp4

### Android
https://user-images.githubusercontent.com/291573/158255555-3de5c3e2-fd39-43fd-b163-4797747620f5.mp4

## How Has This Been Tested?
An animation showcase screen was added in the sample app.

## Types of changes

### CODE REVIEW

**Comment by martingagnon** [comment]
Still reading the code, :amazed:

**Review by martingagnon — COMMENTED** [review]
I did not work with Declarative VMs yet so I cannot give a profound opinion.

From my point of view, its 👍 . I think the WithAnimation block should be the way to go.

**Comment by antoinelamy** [comment]
> I did not work with Declarative VMs yet so I cannot give a profound opinion.
> 
> From my point of view, its 👍 . I think the WithAnimation block should be the way to go.

Not sure if it means the way to go over the other method or not but both the `withAnimation` block and `bindXAnimated` functions are required because the block based API cannot properly enclose a change in a reactive chain.

File: trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/extensions/AnimationExtensions.kt [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+import androidx.compose.animation.core.LinearOutSlowInEasing
+import androidx.compose.animation.core.snap
+import androidx.compose.animation.core.spring
+import androidx.compose.animation.core.tween
+import com.mirego.trikot.viewmodels.declarative.animation.VMDAnimation
+import com.mirego.trikot.viewmodels.declarative.animation.VMDAnimationEasing
+import kotlin.time.DurationUnit
+
+fun <T, V> VMDAnimatedPropertyChange<T, V>.animationSpec() : AnimationSpec<T> {
+    return propertyChange.animation?.toComposeAnimationSpec() ?: snap()
```
npresseault, line 18: I'd go for the = syntax instead of return for the 3 methods here

PR #48, opened 2022-03-15
[Kword] Dynamic I18N and support multiple json paths

npresseault:
## Description
- Introduce a new `DynamicI18N` concept which expose strings via Publishers. It is implemented via a `MultiLanguageI18N` to react on language changes. When the language switches all publishers will emit the localized string in the new language.

- Allow loading json strings from multiple sources (paths) via the `basePaths` parameter.

- Remove code duplication between platform string loaders by moving part of the logic to `KwordLoader` in the common code.

- Test utils from **Trikot Streams** project was extracted so they can be used from outside of it.

## Motivation and Context
- Dynamic i18N was required as we wanted to switch strings without reloading the view.
- `basePaths` was required as we needed to have multiple translations files to be merged into a single i18N and allow overrides.

## How Has This Been Tested?
This has been living in part as an separate library for quite a while for now. The refactor was also tested in externally.

## Types of changes


trikot-kword/kword/src/commonMain/kotlin/com/mirego/trikot/kword/DynamicI18N.kt
```diff
@@ -0,0 +1,15 @@
+package com.mirego.trikot.kword
+
+import com.mirego.trikot.streams.reactive.BehaviorSubject
+import org.reactivestreams.Publisher
+
+interface DynamicI18N {
+    val language: BehaviorSubject<String>
+    val i18N: Publisher<I18N>
```
Boubalou, line 8: Why is the N in caps?

trikot-kword/kword/src/jsMain/kotlin/com.mirego.trikot.kword.js/JsKWord.kt
```diff
@@ -1,39 +1,22 @@
 package com.mirego.trikot.kword.js
 
 import com.mirego.trikot.kword.I18N
-import com.mirego.trikot.kword.KWord
-import kotlinext.js.require
+import com.mirego.trikot.kword.KwordLoader
 
 object JsKWord {
+    private val DEFAULT_PATHS = listOf(
+        "translation"
```
Boubalou, line 8: You don't need the sub-path `translations` for the Web?

Boubalou (approved):

trikot-kword/kword/src/commonMain/kotlin/com/mirego/trikot/kword/DynamicI18N.kt
Boubalou: i18n == `Internationalization`, I don't see the N as a new word. :)

trikot-kword/kword/src/jsMain/kotlin/com.mirego.trikot.kword.js/JsKWord.kt
Boubalou: You are referring to this? `require("./translations/$path")`.  Would it not work if you changed it to `require(".$path")` and use `/translations/translation` as the given default?

npresseault: exactly, the js compiler extracts the `./translations` part of the string literal and passes it to webpack to include the resources under that folder.

trikot-kword/kword/src/commonMain/kotlin/com/mirego/trikot/kword/KwordLoader.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    fun setCurrentLanguageCodes(i18N: I18N, basePaths: List<String>, vararg codes: String) {
+        val map = mutableMapOf<String, String>()
+        basePaths.forEach { basePath ->
+            val variant = mutableListOf<String>()
+            codes.forEach { code ->
+                variant.add(code)
+                val variantPath = variant.joinToString(".")
+                val filePath = "$basePath.$variantPath.json"
+                PlatformTranslationLoader.loadTranslations(filePath)
```
antoinelamy, line 22: We load multiple translation files even if they are not needed?

PR #50, opened 2022-03-15
[Streams] Coroutines interop with Publishers

npresseault:
## Description
Provide coroutine interop with Publishers
This is a fork of https://github.com/Kotlin/kotlinx.coroutines/blob/master/reactive/kotlinx-coroutines-reactive/ with the following modifications:
- It is now published as a KMM module, except for js target where there are some unbound symbols.
- Removed JVM specific code and Reactor lib specific code.
- Removed deprecated functions

## Motivation and Context
This should serve as a bi-directional bridge between our reactive publishers and coroutine flows and channels.

## How Has This Been Tested?
It hasn't been tested yet and will be a bit later.
Unit tests from original codebase weren't ported either.

## Types of changes


martingagnon (approved):
LGTM! Tests would have been nice to confirm at least that logic works.

npresseault:
Importing unit tests from kotlin.coroutines was really not trivial as they were made for JVM target and has quite a few threading logic tests. Those tests also depend on internal test classes that I had to fork as well and adapt for our MPP structure.

That being said, I got them somewhat working with some work.

PR #53, opened 2022-03-18
[VM] Make UIButton extension positionSubviews overridable

charleyxiv:
## Description
UIButton subclasses can now override `positionSubviews` to make further adjustments once layout was adjusted to respect the image's alignment.
 
## How Has This Been Tested?

I tested that a subclass of `UIButton` could override `positionSubviews`:
```
private class Test: UIButton {

    override func positionSubviews(_ alignment: Alignment?) {
        super.positionSubviews(alignment)
    }
}
```

## Types of changes

### DIFF
--- a/trikot-viewmodels/swift-extensions/UIButtonExtensions.swift
+++ b/trikot-viewmodels/swift-extensions/UIButtonExtensions.swift
@@ -156,7 +156,8 @@ extension UIButton {
         return completedAttributedString
     }
 
-    public func positionSubviews(_ alignment: Alignment?) {
+    @objc
+    open func positionSubviews(_ alignment: Alignment?) {
         guard let buttonViewModel = buttonViewModel else { return }
         guard let alignment = alignment, [Alignment.right, Alignment.left].contains(alignment) else { resetAlignment() ; return }
 


marcantoinefortier (approved):

trikot-viewmodels/swift-extensions/UIButtonExtensions.swift
```diff
@@ -156,7 +156,8 @@ extension UIButton {
         return completedAttributedString
     }
 
-    public func positionSubviews(_ alignment: Alignment?) {
+    @objc
```
marcantoinefortier, line 159: Why is this required?

charleyxiv: I needed it to fix this error:

> Non-@objc instance method 'positionSubviews' is declared in extension of 'UIButton' and cannot be overridden

antoinelamy: Maybe a more elegant way of achieving the same result would have been to conditionally check for a specific protocol conformance instead of overriding a function from an extension.

PR #54, opened 2022-03-18
[Streams] Introduce collect() operator for Publishers

marcantoinefortier:
## Description

Collects items emitted by the source Publisher into a single data structure and returns a Publisher that emits this updated structure.

Marbles diagram

## Motivation and Context

I had a use case in my app where I needed such an operator! 😄 

## How Has This Been Tested?

This has been fully unit tested and used in a production app here at Mirego

## Types of changes

### DIFF
--- a/trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/processors/CollectProcessor.kt
+++ b/trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/processors/CollectProcessor.kt
@@ -1,0 +1,39 @@
+import com.mirego.trikot.foundation.concurrent.atomic
+import com.mirego.trikot.streams.reactive.StreamsProcessorException
+import com.mirego.trikot.streams.reactive.processors.AbstractProcessor
+import com.mirego.trikot.streams.reactive.processors.ProcessorSubscription
+import org.reactivestreams.Publisher
+import org.reactivestreams.Subscriber
+
+typealias CollectorBlock<T, R> = (R, T) -> R
+
+class CollectProcessor<T, R>(
+    parentPublisher: Publisher<T>,
+    private val seed: R,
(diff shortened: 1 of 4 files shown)


trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/processors/CollectProcessor.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.mirego.trikot.streams.reactive.processors.AbstractProcessor
+import com.mirego.trikot.streams.reactive.processors.ProcessorSubscription
+import org.reactivestreams.Publisher
+import org.reactivestreams.Subscriber
+
+typealias CollectorBlock<T, R> = (R, T) -> R
+
+class CollectProcessor<T, R>(
+    parentPublisher: Publisher<T>,
+    private val seed: R,
```
npresseault, line 12: I think we need to provide a supplier, not the value itself.
`private val supplier: () -> R`

The issue here is that the value is kept at the processor level, while it should be created during subscription and passed to the subscription. Each new subscription should have a new seed created.

trikot-streams/streams/src/commonTest/kotlin/com/mirego/trikot/streams/reactive/processors/CollectProcessorTests.kt
```diff
@@ -0,0 +1,161 @@
...
+import com.mirego.trikot.streams.reactive.collect
+import com.mirego.trikot.streams.reactive.subscribe
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
+
+class CollectProcessorTests {
+    @Test
+    fun testCollectToList() {
```
npresseault, line 15: we'd need a test for multiple subscriptions.

PR #56, opened 2022-03-22
[Streams] Rename Collect processor which is actually ScanWith

marcantoinefortier:
## Description

Rename our newly added `Collect` processor to `ScanWith` as the current implementation actually responds to the [Scan](https://reactivex.io/documentation/operators/scan.html) reactive spec, not [Reduce/Collect](https://reactivex.io/documentation/operators/reduce.html).

Screen Shot 2022-03-21 at 16 22 32

## Types of changes

This is a **breaking change** as the `Publisher.collect` extension is removed and because the `Publisher.scan` operator now properly dispatches the provided initial value on subscription.

This partially reverts changes of previous commit from https://github.com/mirego/trikot/pull/54

### DIFF
--- a/trikot-streams/streams/src/commonTest/kotlin/com/mirego/trikot/streams/reactive/processors/ScanWithProcessorTests.kt
+++ b/trikot-streams/streams/src/commonTest/kotlin/com/mirego/trikot/streams/reactive/processors/ScanWithProcessorTests.kt
@@ -173,7 +173,6 @@ class ScanWithProcessorTests {
     fun testThrowingSupplier() {
         val publisher = Publishers.behaviorSubject("a")
 
-        @Suppress("ASSIGNED_BUT_NEVER_ACCESSED_VARIABLE")
         var receivedException: StreamsProcessorException? = null
 
         val scanPublisher = publisher.scanWith({ throw IllegalStateException() }) { acc, _ -> acc }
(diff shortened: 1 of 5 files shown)


trikot-streams/streams/src/commonTest/kotlin/com/mirego/trikot/streams/reactive/processors/ScanWithProcessorTests.kt
```diff
@@ -122,6 +169,25 @@ class CollectProcessorTests {
...
+        @Suppress("ASSIGNED_BUT_NEVER_ACCESSED_VARIABLE")
+        var receivedException: StreamsProcessorException? = null
+
+        val scanPublisher = publisher.scanWith({ throw IllegalStateException() }) { acc, _ -> acc }
+
+        assertFailsWith(IllegalStateException::class) {
+            scanPublisher.subscribe(
+                CancellableManager(),
+                onNext = { },
+                onError = { receivedException = it as StreamsProcessorException }
```
npresseault, line 185: You should assert instead

Maybe you could use one of the 2 following functions?
```

inline fun <T : Any, reified R : T> assertIsInstance(actual: T?, message: String? = null): R {
    assertTrue(actual is R, message)
    return actual
}

inline fun <reified R> Any.assertThisIsInstance(message: String? = null): R {
    assertTrue(this is R, message)
    return this
}
```

npresseault (approved):

PR #58, opened 2022-04-14
Expose downloadImageAdapter to DefaultImageViewModelHandler

etienneasselin:
## Description

We expose the downloadImageAdapter to the DefaultImageViewModelHandler so we can override the default behaviour.

## Motivation and Context

We need to handle SVG image type, unfortunately UIKit does not handle SVG on is own. 
`UIImage(data: data)` with svg as data returns nil.
The solution is to expose a method `downloadImageAdapter` to let the applications handle the data itself.

## How Has This Been Tested?

The default behaviour remains the same and was tested here https://github.com/mirego/bellmedia-sports-mobile/pull/3654

## Types of changes


marcantoinefortier (approved):
Could we do the same for the KingFisherImageViewModelHandler?

PR #61, opened 2022-04-20
[Build cleanup] Remove targets duplication

npresseault:
## Description
Remove some duplication when declaring the same KMM targets in several modules

## Motivation and Context
Just some cleanup, re-defining the same targets over and over.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -3,5 +3,11 @@ plugins {
 }
 
 repositories {
+    google()
     mavenCentral()
+}
+
+dependencies {
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.10")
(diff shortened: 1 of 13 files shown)


buildSrc/build.gradle.kts
```diff
@@ -3,5 +3,11 @@ plugins {
 }
 
 repositories {
+    google()
     mavenCentral()
 }
+
+dependencies {
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.10")
```
marcantoinefortier, line 11: Can we get the kotlin version from our gradle.properties here?

marcantoinefortier (approved):

buildSrc/build.gradle.kts
npresseault: Our versions are defined in kotlin in Versions file. This buildSrc gradle script is resolved earlier than that (obviously).

PR #65, opened 2022-04-27
[VMD] Add VMDLazyColumn and VMDLazyRow compose components + update android sample

js-bonin:
## Description
I added the following compose components for VMDListViewModel : VMDLazyColumn, VMDLazyColumnIndexed, VMDLazyRow and VMDLazyRowIndexed.

## Motivation and Context
We have a generic VMDList component in compose that can only layout the list vertically, I think the naming is too generic because it assume we want to layout in a vertical list. I added VMDLazyColumn, VMDLazyColumnIndexed, VMDLazyRow and VMDLazyRowIndexed to allow layout in row or column and to have the ability to receive the index for example if we want to do a zebra list or add divider except for the last item etc.

## How Has This Been Tested?
I added a list showcase screen in the sample which uses those new components

## Types of changes
- [x ] New feature (non-breaking change which adds functionality)

Screenshot_20220427_104627

### DIFF
--- a/trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDLazyColumn.kt
+++ b/trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDLazyColumn.kt
@@ -6,9 +6,8 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyItemScope
-import androidx.compose.foundation.lazy.LazyListScope
 import androidx.compose.foundation.lazy.LazyListState
-import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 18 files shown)


trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDList.kt
```diff
@@ -21,6 +21,7 @@ import com.mirego.trikot.viewmodels.declarative.compose.extensions.isOverridingA
 import com.mirego.trikot.viewmodels.declarative.compose.extensions.observeAsState
 import com.mirego.trikot.viewmodels.declarative.content.VMDIdentifiableContent
 
+@Deprecated("Use either VMDLazyColumn or VMDLazyRow instead")
```
npresseault, line 24: I would say **Use VMDLazyColumn** as this was backed by a LazyColumn and did not support horizontal anyways.

trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDLazyColumn.kt
```diff
@@ -0,0 +1,78 @@
...
+        contentPadding = contentPadding,
+        reverseLayout = reverseLayout,
+        verticalArrangement = verticalArrangement,
+        horizontalAlignment = horizontalAlignment,
+        flingBehavior = flingBehavior
+    ) {
+        itemsIndexed(listViewModel.elements, key = { _, item -> item.identifier }, itemContent = itemContent)
+    }
+}
+
```
npresseault, line 77: extra spacing here

PR #69, opened 2022-05-30
[KWORD] Added an implementation of I18N that works with the Jetpack Compose Previewer.

gbourassa:
## Description

The default implementation of Android kword does not work properly in the context of the Jetpack Compose previewer (it cannot load the file since it's not running the in the same context than the actual app).

## Motivation and Context

Having previews with the actual translations is useful.

## How Has This Been Tested?

In a project.

## Types of changes

### DIFF
--- a/trikot-kword/kword/src/androidMain/kotlin/com/mirego/trikot/kword/android/PreviewI18N.kt
+++ b/trikot-kword/kword/src/androidMain/kotlin/com/mirego/trikot/kword/android/PreviewI18N.kt
@@ -1,0 +1,49 @@
+package com.mirego.trikot.kword.android
+
+import com.mirego.trikot.kword.DefaultI18N
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.json.Json
+import java.io.File
+
+/**
+ * When using the Jetpack Compose previewer in Android Studio, it may be useful to have the actual translations. Since it's
+ * running in a different context than the app, the default Android implementation of I18N is not working.
+ *
+ * This implementation loads a translation file using an absolute path to the translations file in the developer environment.
+ *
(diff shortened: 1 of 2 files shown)


sdetilly (approved):
Nice! Ce serait le fun d'avoir cette info dans un README quelque part pour que cette feature soit plus évidente

trikot-kword/kword/src/androidMain/kotlin/com/mirego/trikot/kword/android/PreviewI18N.kt
```diff
@@ -0,0 +1,49 @@
...
+
+ * First, add an entry to the non version-controlled local.properties:
+ * ```
+ * translation_file_path=/absolute/path/to/translation/file.json
+ * ```
+ *
+ * Then, add this snippet to your build.gradle file (example in Kotlin)
+ * ```
+ * val properties = gradleLocalProperties(rootDir)
+ * val translationFilePath: String = properties.getProperty("translation_file_path", "")
```
npresseault, line 24: I feel we could do better than that and resolve the absolute path using gradle project root or root dir?
```
project.rootDir
project.projectDir
```

npresseault (approved):

trikot-kword/kword/src/androidMain/kotlin/com/mirego/trikot/kword/android/PreviewI18N.kt
gbourassa: Hmm, maybe. I've tried many different ways to get this information at runtime in the Previewer context, without success. But you're right, at the gradle level it would be easier to get it.

gbourassa: @npresseault This was a really good idea, thanks! 👍

PR #71, opened 2022-06-02
Separate kword packages and add support for flows

sdetilly:
## Description
This PR has 2 main goals: 

## First: We add Flows support
This adds support to use Kword multilingual functionalities with Kotlin Coroutines

## Second: We separate the Kword library into 3 packages: 
- The main one (core functionalities, no multilingual support)
- Streams (multilingual support with Publishers)
- Flows (multilingual support with Flows)

Both streams and flows packages can be used at the same time, for projects that would like to migrate from streams to flow for example.

## Motivation and Context

The motivation for this change is to enable support for projects using the new memory model with coroutines.

## How Has This Been Tested?

You may see the kword/sample directory for an example on how it works

## Types of changes

The breaking change is only in the dependency name. For projects already using Multilingual features with publishers, you will need to import an additional dependency: `trikot-kword:kword-streams`. 

Here is a video of the sample:

https://user-images.githubusercontent.com/12590258/171639701-e2fda7db-dd72-4e7f-ab2e-aaba1e0d732b.mp4


trikot-kword/sample/ios/Sample/Sample/SampleApp.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  SampleApp.swift
+//  Sample
```
npresseault, line 3: remove the header

trikot-kword/sample/ios/Sample/SampleApp.swift
```diff
@@ -0,0 +1,25 @@
...
+    var body: some Scene {
+        let serviceLocator = ServiceLocatorImpl(i18N: createI18NMap())
+        WindowGroup {
+            ContentView(textProvider: serviceLocator.textProvider)
+        }
+    }
+    
+    private func createI18NMap() -> FlowMultiLanguageI18N {
+        let i18NEn = DefaultI18N()
+        KwordLoader.shared.setCurrentLanguageCode(i18N: i18NEn, basePaths: ["translation"], code: "en")
```
npresseault, line 15: basePaths should be a const and we could refactor this into a for (en, fr) loop

marcantoinefortier (commented):
Late 👍🏻

PR #74, opened 2022-06-17
[VMD] Format, unformat and transform text

philippelamy1:
## Description

We add an `unformatText` and `transformText` function and we fix the bindings for `formatText` (iOS)

## Motivation and Context

In our project we need to format a textfield value into phone number. We use `formatText` to format the value, `unformatText` to make sure only digits are passed to common, and `transformText` to limit the number of characters. The formatText in VMDTextfield (swift) wasn't working and we found out that since iOS 15 "it might be impossible to manipulate the binding itself" (source: https://itnext.io/swiftui-limit-the-amount-of-characters-in-a-textfield-view-906e819529a5)

So we fix this by replacing `text` by a `@State var` and we use the function `.onChange` to apply the transformations.

## How Has This Been Tested?

Directly into our project:

https://user-images.githubusercontent.com/71347410/174296366-5305598e-650f-4d28-a84b-d50841958f05.mp4

## Types of changes


trikot-kword/sample/common/TRIKOT_FRAMEWORK_NAME.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'TRIKOT_FRAMEWORK_NAME'
-    spec.version                  = '3.3.4'
+    spec.version                  = '3.3.5'
```
remi, line 3: I don’t think the version should be bumped _inside_ the feature PR.

hugolefrancois, line 3: I think that this number will be incremented automatically by the release script? @npresseault

trikot-viewmodels-declarative/sample/common/TRIKOT_FRAMEWORK_NAME.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'TRIKOT_FRAMEWORK_NAME'
-    spec.version                  = '3.3.4'
+    spec.version                  = '3.3.5'
```
hugolefrancois, line 3: I think that this number will be incremented automatically by the release script? @npresseault

hugolefrancois (approved):

trikot-kword/sample/common/TRIKOT_FRAMEWORK_NAME.podspec
npresseault: It is, the file is automatically generated. I wonder if we should keep it under source control.

trikot-viewmodels-declarative/swift/swiftui/Components/VMDTextField.swift
```diff
@@ -12,13 +12,7 @@ public struct VMDTextField<Label>: View where Label: View {
...
     }
 
-    private var text: Binding<String> {
-        Binding {
-            self.viewModel.text
-        } set: { text in
-            self.viewModel.onValueChange(text: self.viewModel.formatText(text))
-        }
-    }
+    @State private var text: String = ""
```
antoinelamy, line 15: Really not sure about that change, any text transformation should be handled by the view model anyway.

PR #81, opened 2022-07-06
Setup GitHub actions & add CD pipeline

npresseault:
## Description
This setups github action with 2 new jobs
- CD which will run when merging in master and build dev versions using the `X.X.X-devXXXX` format
- Release which will run standard release (previously was done on jenkins)

## Motivation and Context
- First motivation was to move from jenkins to github actions.
- Second motivation was to setup a CD pipeline and release dev versions.

## How Has This Been Tested?
Tested some parts of the job using [act](https://github.com/nektos/act) but since it only has ubuntu images, I couldn't test all of it.

## Types of changes
Build/release pipeline improvements


remi (approved):
😲 🙌

PR #83, opened 2022-07-06
Remove freezing tests

npresseault:
## Description
Delete freezing related tests as we are now moving away from the old memory model.
Those tests weren't passing on CD as this property is set: `kotlin.native.binary.freezing=disabled`

## Motivation and Context
CD was failing due to native tests failed with freezing deactivated.

## How Has This Been Tested?
We'll see in next CD run

## Types of changes


thermech (commented):
👍

PR #92, opened 2022-09-14
Remove unavailable stored properties

etienneasselin:
## Description

We fix Xcode 14.0 compilation error
*Stored properties cannot be marked potentially unavailable with '@available'*

## Types of changes


thermech (approved):

antoinelamy:
I think we forgot to apply this change to `trikot-viewmodels-declarative-flow` as well. Opened #93

PR #96, opened 2022-09-28
[VMD-flow] Fix ButtonViewModel setAction staying subscribed

npresseault:
## Description
Fix an issue with setAction with a flow where it would stay subscribed to the flow and potentially re-trigger the action.
VMD (publishers) dealt with this by using first() so similarly we do get the first value (if available) and call the action with it.

## Motivation and Context
When migrating our project from VMD to VMD-flow, this bug was found and it had the side effect to re-trigger our actions multiple times.

## How Has This Been Tested?
It no longer stay subscribed after the fix.

## Types of changes


trikot-viewmodels-declarative-flow/viewmodels-declarative-flow/src/commonMain/kotlin/com/mirego/trikot/viewmodels/declarative/components/impl/VMDButtonViewModelImpl.kt
```diff
@@ -29,7 +31,8 @@ open class VMDButtonViewModelImpl<C : VMDContent>(
     fun <T> setAction(flow: Flow<T>, action: (T) -> Unit) {
         actionBlock = {
             coroutineScope.launch {
-                flow.collect { action(it) }
+                val value = flow.firstOrNull()
+                value?.let { action(it) }
```
antoinelamy, line 35: Inline maybe?
`flow.firstOrNull()?.let { action(it) }`

antoinelamy (approved):
Good catch 👍🏻

PR #98, opened 2022-10-02
Add distinctAppId to the AnalyticsService

MathieuLegault1:
## Description

This PR add a new function on the AnalyticsService that expose the distinct id that the external provider uses to identify a device. It needs to be exposed as a `Promise` since the Android Firebase implementation expose the property as a async Task.

## Motivation and Context

Some of `Mixpanel` APIs requires the distinctId.

## How Has This Been Tested?

This has been tested in my project.

## Types of changes


trikot-analytics/analytics/src/commonMain/kotlin/com/mirego/trikot/analytics/AnalyticsService.kt
```diff
@@ -16,6 +17,11 @@ interface AnalyticsService {
      */
     var isEnabled: Boolean
 
+    /*
+    The distinctAppId associated with the current device
+     */
+    fun distinctAppId(): Promise<String?>
```
gbourassa, line 23: Does the String really need to be nullable? If it's null, shouldn't the  Promise be rejected instead?

gbourassa (approved):

PR #99, opened 2022-10-04
Prevent crashing on too large bitmap in VMDImage on Android

mathieularue:
## Description
We add a check on the bitmap size too prevent crashing later in the drawing process if the size exceeds the limit set by Android.

## Motivation and Context
When downloading remote images there was no check on the resulting drawable size, so anywhere using remote images it could crash due to this limitation.

## How Has This Been Tested?
This has been tested in an ongoing development app that we encountered the issue.

## Types of changes

### DIFF
--- a/trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDImage.kt
+++ b/trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDImage.kt
@@ -145,8 +145,7 @@ fun RemoteImage(
     contentScale: ContentScale = ContentScale.Fit,
     colorFilter: ColorFilter? = null,
     contentDescription: String = "",
-    allowHardware: Boolean = true,
-    displayPlaceholderOnError: Boolean = true
+    allowHardware: Boolean = true
 ) {
     val coilPainter = rememberAsyncImagePainter(
         ImageRequest.Builder(LocalContext.current)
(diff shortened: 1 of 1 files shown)


trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDImage.kt
```diff
@@ -150,16 +156,27 @@ fun RemoteImage(
...
+                )
+            } else {
+                Log.e(TAG, "Unable to load bitmap: size too large (${drawable.bitmap.allocationByteCount})")
+
+                if (displayPlaceholderOnError) {
+                    placeholder(placeholderImage, state)
+                }
+            }
+        }
+        else -> placeholder(placeholderImage, state)
```
gbourassa, line 179: This "else" is an error too. Maybe it should use the displayPlaceholderOnError flag too? Or maybe this flag is useless and the new case should always use the placeholder, just like for "network errors" ?

gbourassa (approved):

trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDImage.kt
mathieularue: True we can call directly placeholder() and if the user doesn't provide one it will fallback to an empty composable anyway, which is the same behaviour as this flag

npresseault:
Can you also apply this fix to VMD-Flow?

PR #100, opened 2022-10-04
Prevent crashing on too large bitmap in VMDImage on Android

mathieularue:
## Description
This is the exact same fix as https://github.com/mirego/trikot/pull/99, but for flow!

## Types of changes


sdetilly (approved):
Thank you for doing this on both libs 👍

PR #103, opened 2022-10-17
[New Module] Added VMD Compiler plugin

gbourassa:
## Description

This new module is a Gradle plugin that uses annotation processing to generate the boilerplate code needed when using Trikot ViewModel Declarative (or its -flow variant).

## Motivation and Context

Boilerplate code writing should be kept to a minimum. Code generation is sometime the solution.

Adding a simple `@Published` annotation to a view model field will automatically generate a base class that contains all the code needed to make this field properly "published".

## How Has This Been Tested?

The "stream" version of the plugin is used in a production project and has been battle tested. The Flow variation is more recent, and should be considered as "early alpha", since no project has actively used it yet.

## Types of changes


martingagnon (approved):
LGTM! 👏

npresseault:
I feel we would all benefit from a few unit tests against generated code.
https://github.com/tschuchortdev/kotlin-compile-testing could help you get those running

gbourassa:
> I feel we would all benefit from a few unit tests against generated code. https://github.com/tschuchortdev/kotlin-compile-testing could help you get those running

@npresseault I totally agree, but I lacked time to add those while extracting the lib from the project in which it was originally developed.
I'll definitly come back later to add some tests, it's in my short-term todo list.
(thanks for the link, that looks interesting, since testing code generation is not always straighforward)

PR #113, opened 2022-11-09
Added step in CD workflow to add a tag when publishing a new DEV version (needed for cocoapod)

gbourassa:
## Description

In the current CD workflow, a new dev version is built and published for each new commit in master branch. However, this would only release the "maven" artefacts. For project using cocoapod, the new release would not be available since it requires that a tag is added on github.
This adds an extra step to properly create and push the tag.

## Motivation and Context

The usual flow was manually create a tag each time a new releases was needed. Now it is automatic.

## Types of changes

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -45,3 +45,4 @@ jobs:
         run: |
           ./gradlew writeDevVersion -i -s
           ./gradlew check publish -i -s
+          ./gradlew tagDevVersion -i -s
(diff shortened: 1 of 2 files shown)


npresseault (approved):

antoinelamy:
As a stretch goal we could also properly publish versions to CocoaPods repository. We would have to validate the podspec with `pod spec lint` and publish it with `pod trunk push`.

PR #125, opened 2022-11-22
Add a DSL in order to ease view model creation

mathieularue:
## Description
This provides a DSL in order to create common VM components, we can use it as follows:
```
text("My label")
localImage(ImageResource.MY_IMAGE)
buttonWithText("My button" action = {}) 
```

Instead of:
```
VMDComponents.Text.withContent("My label", cancellableManager)
VMDComponents.Image.local(ImageResource.MY_IMAGE, cancellableManager)
VMDComponents.Button.withText("My button", cancellableManager)
```

## Motivation and Context
Most of the time when creating view models we are already inside a VMDViewModelImpl context, so we should be able to create components view models easily without passing the cancellableManager  / coroutineScope everywhere. Using an interface we can easily extend the DSL with custom definitions inside each projects if needed. Also if we are in the context of a "content" class or anything other than a view model we just need to implement the interface in order to be able to use the DSL.

## How Has This Been Tested?
This has been tested locally in other projects

## Types of changes


marcantoinefortier (approved):
That'd be great if the samples used this new DSL :)

mathieularue:
@marcantoinefortier this surely can be done, on it!

npresseault:
I think we're missing the newest picker component here! But overall looks clean

PR #129, opened 2022-12-02
Remove publish gradle plugin

npresseault:
## Description
Stop using the release plugin, use publish and tag instead

## Motivation and Context
We were previously using the release plugin which was bumping versions automatically in git.
With the new release process, we want to manage the bumping of versions manually instead.

## How Has This Been Tested?
Not at all, we'll test after it's merged !

## Types of changes

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -45,4 +45,4 @@ jobs:
         run: |
           ./gradlew writeDevVersion -i -s
           ./gradlew check publish -i -s
-          ./gradlew tagDevVersion -i -s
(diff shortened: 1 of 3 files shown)


remi (approved):

gbourassa (approved):

Boubalou (approved):
With this change, do I need to revert the change to the version `4.0.0-dev1->4.0.0` I manually made? (and remove the Tag as well)

npresseault:
The tag yes, it will be created at the end of the build. The version must be set to X.X.X without the -dev suffix during the release otherwise we could create a task that removes it for the build.

PR #131, opened 2022-12-09
[VMD-flow] Use of the viewModelScope on Android

antoinelamy:
## Description

VMDViewModelController on Android was managing its own coroutineScope instead of using the lifecycle provided scope.

## Motivation and Context

Ensure that a lifecycle managed scope is used on Android.

## How Has This Been Tested?

In the sample app.

## Types of changes

Breaking change is mostly due to the renaming of the scope, to fix, rename `viewModelControllerScope` to `viewModelScope`. The new name is more consistent with what the scope does, view model controller manages the lifecycle events for the view model, providing a scope to the view model. It's therefore the scope of the view model.

### DIFF
--- a/trikot-viewmodels-declarative-flow/viewmodels-declarative-flow/src/commonMain/kotlin/com/mirego/trikot/viewmodels/declarative/controller/VMDViewModelController.kt
+++ b/trikot-viewmodels-declarative-flow/viewmodels-declarative-flow/src/commonMain/kotlin/com/mirego/trikot/viewmodels/declarative/controller/VMDViewModelController.kt
@@ -3,24 +3,16 @@ package com.mirego.trikot.viewmodels.declarative.controller
 import com.mirego.trikot.foundation.concurrent.atomic
 import com.mirego.trikot.foundation.ref.weakAtomicReference
 import com.mirego.trikot.viewmodels.declarative.viewmodel.VMDViewModel
-import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.SupervisorJob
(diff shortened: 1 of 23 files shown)


trikot-viewmodels-declarative-flow/viewmodels-declarative-flow/src/commonMain/kotlin/com/mirego/trikot/viewmodels/declarative/controller/VMDViewModelController.kt
```diff
@@ -3,25 +3,17 @@ package com.mirego.trikot.viewmodels.declarative.controller
...
 
-    open val exceptionHandler: CoroutineExceptionHandler = CoroutineExceptionHandler { _, throwable ->
-        println("CoroutineExceptionHandler got $throwable")
-    }
-
-    protected val viewModelControllerScope = CoroutineScope(Dispatchers.Main.immediate + SupervisorJob() + exceptionHandler)
-
     abstract val viewModel: VM
 
+    val viewModelScope: CoroutineScope = platformViewModelScope
```
mathieularue, line 15: Do we really need this val? Or we could only have platformViewModelScope renamed to viewModelScope?

mathieularue (approved):

trikot-viewmodels-declarative-flow/viewmodels-declarative-flow/src/commonMain/kotlin/com/mirego/trikot/viewmodels/declarative/controller/VMDViewModelController.kt
antoinelamy: Oddly, if I didn't put it there the compiler is unable to see the expected underlying val.

PR #134, opened 2022-12-15
[VMD-Flow] Compose: added optional support for XML Drawables and 9-patch

gbourassa:
## Description
The default Jetpack Compose `painterResource` only handle vector drawables and rasterized assets. Attempting to use any other type of Drawable will result in this error:

```
java.lang.IllegalArgumentException: Only VectorDrawables and rasterized asset types are supported ex. PNG, JPG
  at androidx.compose.ui.res.PainterResources_androidKt.loadVectorResource(PainterResources.android.kt:94)
  at androidx.compose.ui.res.PainterResources_androidKt.painterResource(PainterResources.android.kt:65)
  at com.mirego.trikot.viewmodels.declarative.compose.extensions.ImageResourceExtensionsKt.painterResource(ImageResourceExtensions.kt:21)
```

Since our `VMDImageProvider` is generic and always uses the default `painterResource`, it is not allowed to map a `VMDImageResource` to an XML drawable or a nine-patch png.

There is now a flag to activate support for these types. It is an opt-in flag (default behavior is like before), since there may be a very slight performance impact, since we have to determine each time if the drawable is a vector or not.

This new feature relies on Google Accompanist library for the actual painter. 

## Implementation
At first look the easy solution to this problem seems to be "catch the Exception thrown and fallback to another painter when this happens. However, it is not possible to catch an Exception in a `@Composable` method. So we _have_ to detect the cases that would eventually throw the Exception before it happens. 

The method to detect resource type is largely inspired (AKA copy/pasted) from the compose `painterResource` code.

## Motivation and Context
A proper, recent declarative-based project usually never need XML drawables. But there are cases where we want to add this UI stack to an old project, and reuse the already existing image mapping. 

## How Has This Been Tested?
In a project that actually contains old XML drawables. But since this is an opt-in flag, behavior will not change for existing projects.

## Types of changes

### DIFF
--- a/trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/extensions/ImageResourceExtensions.kt
+++ b/trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/extensions/ImageResourceExtensions.kt
@@ -1,13 +1,19 @@
 package com.mirego.trikot.viewmodels.declarative.compose.extensions
 
 import android.content.Context
+import android.util.TypedValue
+import androidx.appcompat.content.res.AppCompatResources
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.ImageBitmap
 import androidx.compose.ui.graphics.painter.BitmapPainter
 import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.platform.LocalContext
+import com.google.accompanist.drawablepainter.rememberDrawablePainter
 import com.mirego.trikot.viewmodels.declarative.configuration.TrikotViewModelDeclarative
+import com.mirego.trikot.viewmodels.declarative.configuration.VMDImageProviderConfiguration.painterResourceLegacyDrawableSupport
 import com.mirego.trikot.viewmodels.declarative.properties.VMDImageResource
+import org.xmlpull.v1.XmlPullParser
+import org.xmlpull.v1.XmlPullParserException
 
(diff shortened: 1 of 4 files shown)


trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/extensions/ImageResourceExtensions.kt
```diff
@@ -18,8 +24,53 @@ fun painterResource(imageResource: VMDImageResource?): Painter {
     val context = LocalContext.current
     imageResource?.let { resource ->
         resource.resourceId(context)?.let { id ->
-            return androidx.compose.ui.res.painterResource(id = id)
+            return if (painterResourceLegacyDrawableSupport) {
```
npresseault, line 27: This could be a different function and we would be able to pick the appropriate ones for optimal performance.

npresseault: I feel this should be a flag on the function rather than a global one.

gbourassa: The thing is, `painterResource` is already called from a lot of places, among others in `VMDImage`..  we would need to pass this flag to alot of places.
Additionally, the need to support XML drawables depends on the VMDImageProvider mapping, _not_ on the context where the method is called.

npresseault (approved):

PR #135, opened 2022-12-19
Update to Kotlin 1.7.20

Boubalou:
## Description

Update Trikot to Kotlin `1.7.20`!

We did not move to `1.7.22` due to some missing dependencies around Jetpack Compose that were blocking us, forcing us to split the dependencies into each version because we could not use `1.3.2` based on the RUNTIME as expected.

## Motivation and Context

A few other projects using Trikot were already using `1.7.20` with great success. It's our time to shine and move forward as well! 

## How Has This Been Tested?

It got integrated as a Maven local dependency within our project that uses almost all platforms, including Javascript.

## Types of changes


marcantoinefortier (approved):

marcantoinefortier:
You can also remove [this line](https://github.com/mirego/trikot/blob/512d468c98b828ff21e4edc2e589e70673690e1e/gradle.properties#L13) from the `gradle.properties` as it is now the default value in kotlin 1.7.20

npresseault:
Looks like https://youtrack.jetbrains.com/issue/KT-53183/KAPT-Databinding-compiler-no-longer-resolve-types-in-multiplatform-library-with-kotlin-1.7.0 is fixed, we are good to go!

npresseault (approved):

npresseault:
Dont forget the CHANGELOG!

PR #138, opened 2023-01-03
Disable sample modules via property disable_samples

npresseault:
## Description
Disable samples via a property in CD / Release workflows rather than commenting the modules in source control.

## Motivation and Context
We needed to de-comment those sample modules locally to run them. The flag is more convenient as the default is to enable them and only CD/Release workflows need to disable them.


Boubalou (commented):
Sharp!

PR #139, opened 2023-01-10
Vectorial SwiftUI images can now be scaled

hugolefrancois:
## Description

There is bug in the images mapping in SwiftUI. If you use a vectorial image and you try to scale it up, the image will be blurry. 

This is caused because the images are loaded in an UIImage and we create the SwiftUI Image with that UIImage. The vectorial information of the image is lost in the process 🤷🏼‍♂️ 

Since we can't have the type SwiftUI.Image in the Kotlin code in iosMain, and the interface defined in the common code was only used by the natif code, I have move that in the Swift code on iOS and in the compose files for Android.

⚠️ This will create a breaking change on all iOS projects using SwiftUI. The fix is super easy, you have to build Image directly instead of UImage in your ImageProvider.

## How Has This Been Tested?

Tested in the sample app

## Types of changes


gbourassa (approved):

dominic-plouffe (approved):
LGTM

marcantoinefortier (approved):

PR #140, opened 2023-01-12
iOS: Add VMDForEach and externalActionHandler on VMDButton

hugolefrancois:
## Description

**VMDForEach**
Useful when you want to display a VMDListViewModel in SwiftUI but you don't want to put it in a List (VMDList)

**New VMDButton feature**
Sometime we need to trigger some native code when a VMDButton action is triggered. The new externalActionHandler allow the caller, when the button is tapped, to do stuff before or after the VMDButton action.

PS: Modification done in VMD and VMD-flow

## Motivation and Context

I was using these 2 features in all of my projects so why not moins that in Trikot.

## Types of changes

### DIFF
--- a/trikot-viewmodels-declarative-flow/swift/swiftui/Components/VMDButton.swift
+++ b/trikot-viewmodels-declarative-flow/swift/swiftui/Components/VMDButton.swift
@@ -4,6 +4,8 @@ import TRIKOT_FRAMEWORK_NAME
 public struct VMDButton<Label, Content: VMDContent>: View where Label: View {
     private let labelBuilder: (Content) -> Label
 
+    var externalActionHandler: ((_ action: () -> Void) -> Void)?
+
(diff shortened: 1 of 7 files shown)


trikot-viewmodels-declarative-flow/swift/swiftui/Components/VMDButton.swift
```diff
@@ -16,12 +18,26 @@ public struct VMDButton<Label, Content: VMDContent>: View where Label: View {
...
-        })
-            .disabled(!viewModel.isEnabled)
-            .hidden(viewModel.isHidden)
+        }
+        .disabled(!viewModel.isEnabled)
+        .hidden(viewModel.isHidden)
+    }
+}
+
+extension VMDButton {
```
dominic-plouffe, line 37: Maybe move this into its own `VMButton+Configuration` file? Just to make sure to be consistent with other components.

dominic-plouffe (approved):

PR #144, opened 2023-01-26
Fix SharedProcessor interop with Flow

EtienneStPierre:
## Description

I had a case where this flow as publisher below wasn't emitting any values. 
```        
val publisher = flowOf(1)
    .map { n ->
            n
        }
        .asPublisher()
        .shared()

publisher.subscribe(CancellableManager()) {
   //No emmited value
}
```

Our `SharedProcessor` is a subscriber, therefore it should call `request(n: Long)` on its subscription. 

## Motivation and Context

We use a lot of interoperability between flow and publisher, and this was causing several issues with our shared publishers. 

## How Has This Been Tested?

Test has been added into `FlowAsPublisherTest` to make sure the behaviour of the `SharedProcessor` hasn't changed after the interop. 

## Types of changes

### DIFF
--- a/trikot-streams/coroutines-interop/src/commonTest/kotlin/kotlinx/coroutines/reactive/FlowAsPublisherTest.kt
+++ b/trikot-streams/coroutines-interop/src/commonTest/kotlin/kotlinx/coroutines/reactive/FlowAsPublisherTest.kt
@@ -5,11 +5,8 @@
 package kotlinx.coroutines.reactive
 
 import com.mirego.trikot.streams.cancellable.CancellableManager
-import com.mirego.trikot.streams.reactive.Publishers
-import com.mirego.trikot.streams.reactive.map
 import com.mirego.trikot.streams.reactive.shared
 import com.mirego.trikot.streams.reactive.subscribe
-import kotlin.math.exp
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.flow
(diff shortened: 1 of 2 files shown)


trikot-streams/coroutines-interop/src/commonTest/kotlin/kotlinx/coroutines/reactive/FlowAsPublisherTest.kt
```diff
@@ -4,6 +4,12 @@
 
 package kotlinx.coroutines.reactive
 
+import com.mirego.trikot.streams.cancellable.CancellableManager
+import com.mirego.trikot.streams.reactive.Publishers
+import com.mirego.trikot.streams.reactive.map
+import com.mirego.trikot.streams.reactive.shared
+import com.mirego.trikot.streams.reactive.subscribe
+import kotlin.math.exp
```
npresseault, line 12: there are some unused imports here, run ktlint

trikot-streams/coroutines-interop/src/commonTest/kotlin/kotlinx/coroutines/reactive/FlowAsPublisherTest.kt
```diff
@@ -78,6 +91,23 @@ class FlowAsPublisherTest : TestBase() {
         finish(3)
     }
 
+
+    @Test
+    fun testFlowWithSharedResultIsShared() = runTest {
+        val publisher = flowOf(1)
+            .map { n ->
+                expect(1)
```
npresseault, line 99: Run the expect in the subscription (onEvent), not in the map

npresseault (approved):

trikot-streams/coroutines-interop/src/commonTest/kotlin/kotlinx/coroutines/reactive/FlowAsPublisherTest.kt
EtienneStPierre: Actually, I'm testing that we do share the value after a second subscribe instead of reprocessing the value in the map.

npresseault: Ah gotcha, it does make sense!

PR #147, opened 2023-03-02
Update to kotlin 1.8.10 and remove support for old memory model

npresseault:
## Description
This updates to kotlin 1.8.10. In that version, calling freeze now cause compilation errors.
We are now dropping support for old MM

We are also updating various libraries.

Ktlint was updated so this is why there is a lot of code style changes.

## Motivation and Context
Old MM has become obsolete in favor of the new MM.

## How Has This Been Tested?
I tested it in another project to test ABI compatibility.

## Types of changes

Changes are ABI compatible. Old freezing utils were deprecated and will just do nothing. 
This is however breaking in the sense that we drop support for old memory model.

New android lifecycle library also as a breaking change but that's not on us.


buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -1,28 +1,28 @@
...
-    const val JETPACK_COMPOSE_UI_TOOLING = "1.3.2"
+    const val JETPACK_COMPOSE_UI_TOOLING = "1.3.3"
     const val JETPACK_COMPOSE_MATERIAL = "1.3.1"
-    const val JETPACK_COMPOSE_MATERIAL_3 = "1.1.0-alpha03"
+    const val JETPACK_COMPOSE_MATERIAL_3 = "1.1.0-alpha07"
     const val COIL = "2.0.0-rc03"
-    const val KTLINT = "10.3.0"
+    const val KTLINT = "11.2.0"
     const val KOTLINX_SERIALIZATION = "1.4.1"
     const val KOTLINX_COROUTINES = "1.6.3"
```
marcantoinefortier, line 14: I think we should bump to 1.6.4 (there are some important bug fixes)

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -1,28 +1,28 @@
...
     const val JETPACK_COMPOSE_FOUNDATION = "1.3.1"
-    const val JETPACK_COMPOSE_UI_TOOLING = "1.3.2"
+    const val JETPACK_COMPOSE_UI_TOOLING = "1.3.3"
     const val JETPACK_COMPOSE_MATERIAL = "1.3.1"
-    const val JETPACK_COMPOSE_MATERIAL_3 = "1.1.0-alpha03"
+    const val JETPACK_COMPOSE_MATERIAL_3 = "1.1.0-alpha07"
     const val COIL = "2.0.0-rc03"
-    const val KTLINT = "10.3.0"
+    const val KTLINT = "11.2.0"
     const val KOTLINX_SERIALIZATION = "1.4.1"
```
marcantoinefortier, line 13: [Kotlinx.serialization 1.5.0](https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.5.0) is also out with kotlin 1.8.10 support 🙂

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -1,28 +1,28 @@
 object Versions {
-    const val KOTLIN = "1.7.20"
+    const val KOTLIN = "1.8.10"
     const val ATOMIC_FU = "0.18.5"
```
marcantoinefortier, line 3: [Atomixfu 0.20.0](https://github.com/Kotlin/kotlinx-atomicfu/releases/tag/0.20.0) is out with Kotlin 1.8.10 support 🙂

marcantoinefortier (approved):

PR #152, opened 2023-03-27
Rich Text Support [iOS]

laurentboileau:
## Description

Add rich text support to iOS.

## Motivation and Context

This is a frequent requirement on projects. This implementation uses `AttributedString` on iOS 15+ and uses `UILabel` and `UIViewRepresentable` on iOS 13 and 14.

**EDIT: Support on iOS 13 and 14 was removed since view modifiers such as `foregroundColor` among others could not be applied to the underlying UILabel view.** Instead, a plain text version is displayed. View modifiers can be used in this scenario. Since legacy support was removed, the configuration API is now available for rich text.

## How Has This Been Tested?

Rich text is applied in Sample app when running on iOS simulators (iOS 14.5 and iOS 16.2)

Screenshot 2023-03-27 at 4 43 25 PM copy

## Types of changes

## Discussion Topics

The following is a list of topics I would like to discuss during PR review:

- `VMDText` has a a closure based API to configure the underlying `Text` view. Configurations are stored as a list of closures and are applied by calling `reduce`. This works well as long as a single return type (`Text`) is used. Unfortunately, support on versions prior to iOS 15 required that we introduced a `UIRichText` view that wraps a `UILabel`. As a consequence, the configuration API is now unsupported if rich text is detected (`viewModel.spans` is empty).
- `UIRichText` (For iOS 13 and 14) uses the default implementation of [sizeThatFits(_:uiView:context:)](https://developer.apple.com/documentation/swiftui/uiviewrepresentable/sizethatfits(_:uiview:context:)-9ojeu)
- `TrikotViewModelDeclarative` now expects to be initialized with a `VMDSpanStyleProvider` implementation. This will require all TrikotVMD consumers to provide one. A similar situation is already the case with `VMDImageProvider`
- `TrikotViewModelDeclarative` is declared within **VMDImageProvider.swift**. I propose we move it to its own file.

### DIFF
--- a/trikot-viewmodels-declarative-flow/swift/swiftui/ImageResource/VMDImageProvider.swift
+++ b/trikot-viewmodels-declarative-flow/swift/swiftui/ImageResource/VMDImageProvider.swift
@@ -9,39 +9,3 @@ public struct DefaultImageProvider: VMDImageProvider {
     public init() { }
     public func imageForResource(imageResource: VMDImageResource) -> Image? { nil }
 }
-
-public class TrikotViewModelDeclarative {
-
-    public static let shared: TrikotViewModelDeclarative = TrikotViewModelDeclarative()
-
-    var imageProvider: VMDImageProvider {
-        guard let internalImageProvider = internalImageProvider else {
-            fatalError("TrikotViewModelDeclarative must be initialized before use")
-        }
-
-        return internalImageProvider
-    }
-
-    var spanStyleProvider: VMDSpanStyleProvider {
-        guard let internalSpanStyleProvider = internalSpanStyleProvider else {
-            fatalError("TrikotViewModelDeclarative must be initialized before use")
-        }
-
-        return internalSpanStyleProvider
-    }
-
-    private var internalImageProvider: VMDImageProvider?
-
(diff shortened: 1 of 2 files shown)


trikot-viewmodels-declarative-flow/sample/ios/Sample/SampleSpanStyleProvider.swift
```diff
@@ -0,0 +1,15 @@
+import SwiftUI
+import TRIKOT_FRAMEWORK_NAME
+import Trikot
+
+class SampleSpanStyleProvider: VMDSpanStyleProvider {
```
nlangheit, line 5: Should we change it to SampleTextStyleProvider since we changed to VMDTextStyleResource. Same for VMDSPanStyleProvider?

laurentboileau: I hesitated too. My reasoning is that it's only used in a rich text context. The type provides "span styles" for attributed strings formatting, and does so based on a `VMDTextStyleResource` parameter.

martingagnon (approved):
I'll leave the naming to you :) 

Love it!

PR #157, opened 2023-04-06
[VMD-Compiler] When all fields of a viewmodel are annotated with @Published, the generated base class won't be abstract

gbourassa:
This way they can be used as-is, without the need to subclass them.

## Description

There are some cases where it would be convenient to use the generated implementation as-is. Since it's abstract, it must always be extended by a useless subclass.

With this PR, the compiler will not generate the class as being concrete, when possible.

For instance, if one defines the following view model:

```
interface VMDTextViewModel : VMDViewModel {
    @Published
    val text: String
    @Published
    val spans: List<VMDRichTextSpan>
}
```

Then the generated implementation will be very similar to the official `VMDTextViewModelImpl`, usable as-is (by binding its content externally).

## How Has This Been Tested?

In a project.

## Types of changes


trikot-viewmodels-declarative-compiler/viewmodels-declarative-compiler-flow/src/test/kotlin/com/mirego/trikot/viewmodels/declarative/compiler/FlowVMDCodeGeneratorTest.kt
```diff
@@ -50,7 +52,7 @@ class FlowVMDCodeGeneratorTest {
     }
 
     @Test
-    fun `When an interface contains Published annotations base class is generated `() {
+    fun `When an interface contains only Published annotations default class is generated `() {
```
npresseault, line 55:
```suggestion
    fun `When an interface contains only Published annotations open class is generated `() {
```

trikot-viewmodels-declarative-compiler/viewmodels-declarative-compiler-flow/src/test/kotlin/com/mirego/trikot/viewmodels/declarative/compiler/FlowVMDCodeGeneratorTest.kt
```diff
@@ -67,9 +69,65 @@ class FlowVMDCodeGeneratorTest {
         )
         assertEquals(KotlinCompilation.ExitCode.OK, result.exitCode)
 
+        val generatedClass = result.classLoader.loadClass("DefaultViewModel1Impl").kotlin
+
+        generatedClass.isSuperclassOf(VMDViewModelImpl::class)
+        assertFalse(generatedClass.isAbstract)
```
npresseault, line 75: check if the class is open too

trikot-viewmodels-declarative-compiler/viewmodels-declarative-compiler-core/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compiler/VMDScanner.kt
```diff
@@ -30,10 +34,15 @@ class VMDScanner(
...
 
+                val viewModelProperties = getViewModelProperties(vmdCodeGenerator, resolver, viewModelInterface)
+                val viewModelAbstractMethods = getViewModelAbstractMethods(vmdCodeGenerator, resolver, viewModelInterface)
+
                 ViewModelMetaData(
                     viewModelInterface = viewModelInterface,
                     publishedProperty = entry.value.map { it as KSPropertyDeclaration },
-                    superClass = superClassDeclaration
+                    superClass = superClassDeclaration,
+                    allFieldsArePublished = entry.value.size == viewModelProperties.count(),
```
npresseault, line 44: What about properties such as
```
    val someString: String
        get() = "YOLO"
```

This doesn't need to be counted as it does not need to be overriden

npresseault (approved):

trikot-viewmodels-declarative-compiler/viewmodels-declarative-compiler-core/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compiler/VMDScanner.kt
gbourassa: That would be a weird thing to have on a viewmodel interface, but I agree that this case should be covered, just in case.

PR #159, opened 2023-04-25
Lowering Android min SDK to 21

gbourassa:
## Description

Trikot should try to keep minimum android version requirements as low as possible, so that legacy projects that still use 21 as their min SDK can use it.

There is nothing in the code that actually requires a min SDK of 24.

## Motivation and Context

Some older projects still have their min SDK set at 21, and this should not prevent us from including Trikot.

## How Has This Been Tested?

Android samples are still running, all check tasks still pass, lib can be included in a 21+ project with success.

## Types of changes

### DIFF
--- a/trikot-viewmodels-declarative/sample/android/build.gradle.kts
+++ b/trikot-viewmodels-declarative/sample/android/build.gradle.kts
@@ -53,6 +53,12 @@ android {
             resources.srcDir("../common/src/commonMain/resources/")
         }
     }
+
+    lint {
+        abortOnError = true
+        checkReleaseBuilds = true
+        disable.add("NotificationPermission")
+    }
 }
 
 dependencies {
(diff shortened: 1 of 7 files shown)


trikot-viewmodels-declarative/sample/android/build.gradle.kts
```diff
@@ -53,6 +53,12 @@ android {
             resources.srcDir("../common/src/commonMain/resources/")
         }
     }
+
+    lint {
+        abortOnError = true
+        checkReleaseBuilds = true
+        disable.add("NotificationPermission")
```
npresseault, line 60: there is no way to have this on a more granular level?

npresseault (approved):

trikot-viewmodels-declarative/sample/android/build.gradle.kts
gbourassa: Possibly, but it's the sample project(s), so I'd say it's good enough.
For more context, this is a weird linting issue triggered by Picasso following the update of the Gradle plugin. Updating Picasso version _may_ have helped, but it seemed less risky to simply disable the check.

PR #161, opened 2023-05-17
[VMD] Fix observeAsState for instance properties

npresseault:
## Description
The exception thrown was:
`Fatal Exception: java.lang.IllegalArgumentException: Callable expects 0 arguments, but 1 were provided.`

Since #158 it was no longer possible to call `observeAsState` using a `KProperty0` (instance property). We now properly support `KProperty0` (instance property) and `KProperty1` (class property with a receiver argument)

For reference
```
val kPropReceiver: KProperty0<Boolean> = viewModel::isHidden 
val kProp: KProperty1<VMDViewModel, Boolean> = VMDViewModel::isHidden
```

`KProperty2`: extension function with a different receiver will not be supported.

## Motivation and Context
We want to be able to use both property types

## How Has This Been Tested?
Tested with both property types

## Types of changes


trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/extensions/ViewModelExtensions.kt
```diff
@@ -133,3 +135,12 @@ inline fun <T, S> S.asState(
     }
     return state
 }
+
+@Suppress("UNCHECKED_CAST")
+private  fun <T, V> KProperty<V>.callGetter(receiver: T): V =
```
etienneasselin, line 140: two spaces between private and fun

etienneasselin (approved):

trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/extensions/ViewModelExtensions.kt
npresseault: ktlint seems not to run on those modules, I'll fix this.

npresseault: Ok so I setupped ktlint properly, lots of unrelated changes, sorry for that.

trikot-viewmodels-declarative/compose/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/extensions/ViewModelExtensions.kt
```diff
@@ -111,3 +113,11 @@ inline fun <T, S> S.asState(
     }
     return state
 }
+
+@Suppress("UNCHECKED_CAST")
+private fun <T, V> KProperty<V>.callGetter(receiver: T): V =
```
ccipelli-mirego, line 118: 👍

ccipelli-mirego (commented):
🤙

PR #170, opened 2023-08-02
[VMD-Flow] Test identifiers

npresseault:
## Description
Introduce `testIdentifier` as a nullable field in VMDViewModel and apply it to all VMD Components.
Introduce vmdModifier which will contain all base modifiers of VMDViewModel (`hidden` and `testIdentifier` for now).

## Motivation and Context
`accessibilityIdentifier` and `testTag` are modifiers that can be used to retrieve a compose node or swiftui view during ui tests.

## How Has This Been Tested?
In the sample app

## Types of changes

### DIFF
--- a/trikot-viewmodels-declarative-flow/swift/swiftui/ViewModifiers/VMDModifier.swift
+++ b/trikot-viewmodels-declarative-flow/swift/swiftui/ViewModifiers/VMDModifier.swift
@@ -1,0 +1,29 @@
+import SwiftUI
+import TRIKOT_FRAMEWORK_NAME
+
+struct VmdModifier: ViewModifier {
+    var viewModel: VMDViewModel
+    
+    func body(content: Content) -> some View {
+        content
+            .hidden(viewModel.isHidden)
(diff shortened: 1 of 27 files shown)


etienneasselin (approved):
LGTM

trikot-viewmodels-declarative-flow/swift/swiftui/ViewModifiers/VMDModifier.swift
```diff
@@ -0,0 +1,29 @@
+import SwiftUI
+import TRIKOT_FRAMEWORK_NAME
+
+struct VmdModifier: ViewModifier {
```
laurentboileau, line 4: I think separating each operation into separate modifiers might be a better approach.

npresseault: As we talked earlier, the extension is public so it can be called from outside, the struct really doesn't need to.
Separating each operation would result to duplicate the calls everywhere in every component. Changing how they are created and re-using the modifiers might take more time to think through so this can come in another PR if some SwiftUI kingpin wants to tackle that.

PR #178, opened 2023-08-21
[trikot-viewmodels-declarative-flow] Fix crash in FlowExtensions wrap method

frobert11:
## Description

Fix a crash that could occurred on iOS with the FlowExtensions wrap method if the flow value type is nullable.

In order to fix that, we add support for nullable type.

## Motivation and Context

To fix a crash on iOS.

## How Has This Been Tested?

Tested in a project.

## Types of changes


gbourassa (approved):
Can this somehow be a breaking change? 🤔

sdetilly (approved):

frobert11:
> Can this somehow be a breaking change? 🤔

@gbourassa I dont think it is a breaking change as it would crash if used as is with a nullable value on iOS and I didnt see any behaviour change on Android. Now at least, we can handle it correctly 😄

PR #179, opened 2023-08-24
Dispatch Android Lifecycle addObserver and removeObserver on main thread

alarochelle:
## Description

Android `Lifecycle::addObserver` and `Lifecycle::removeObserver` should only be called on the main thread, this PR makes sure that these functions are always called on the main thread.

## Motivation and Context

This was somewhat unclear in Trikot and could cause crashes in apps subscribing to the ApplicationStatePublisher when subscribing/unsubscribing. Trikot should make sure that these functions are called from the main thread.

## How Has This Been Tested?

Tested in my application.

## Types of changes


gbourassa (approved):

marcantoinefortier (approved):

trikot-viewmodels/viewmodels/src/androidMain/kotlin/com/mirego/trikot/viewmodels/lifecycle/ApplicationStatePublisher.kt
```diff
@@ -21,11 +22,17 @@ actual class ApplicationStatePublisher :
 
     override fun onFirstSubscription() {
         super.onFirstSubscription()
-        lifecycle.addObserver(this)
+
+        UIThreadDispatchQueue().dispatch {
```
marcantoinefortier, line 26: We could reuse the same dispatch queue for both methods, but that's fine like that too 🙂

marcantoinefortier (approved):

PR #182, opened 2023-08-31
VMDFlow declaration change and optional values handling on iOS

laurentboileau:
## Description

Tag `5.0.0-dev2660` introduced a breaking change to `VMDFlow` for iOS apps.

The generated Objective-C header for VMDFlow's watchBlock method now always specifies that the value argument is an optional:

```
- (void)watchBlock:(void (^)(T _Nullable, id<PREFIXCloseable>))block __attribute__((swift_name("watch(block:)")));
```

This PR:

- Changes VMDFlow's kotlin signature from `class VMDFlow<T : Any?>` to `class VMDFlow<out T>`
- Fixes trikot-viewmodels-declarative-flow swift files to account for this change.
- **It does not remove the nullability of watched values.**

## Motivation and Context

Fixes compilation issues on iOS apps.

## How Has This Been Tested?

Changes were built locally and tested against a production project.

## Types of changes

### DIFF
--- a/trikot-viewmodels-declarative-flow/swift/swiftui/Components/Snackbar/VMDSnackBarStateObject.swift
+++ b/trikot-viewmodels-declarative-flow/swift/swiftui/Components/Snackbar/VMDSnackBarStateObject.swift
@@ -11,19 +11,23 @@ public class VMDSnackBarStateObject: ObservableObject {
 
     public func startWatching(_ flow: VMDFlow<VMDSnackbarViewData>) {
         flow.watch { [weak self] viewData, closeable in
-            self?.message = viewData?.message ?? ""
-            self?.duration = viewData?.duration.toTimeInterval() ?? VMDSnackbarDuration.short_.toTimeInterval()
-            self?.isVisible = true
-            self?.withDismissAction = viewData?.withDismissAction ?? true
+            guard let self else { return }
 
-            if self?.closeableExecuter == nil {
-                self?.closeableExecuter = VMDCloseableExecuter(closeable: closeable)
+            if self.closeableExecuter == nil {
+                self.closeableExecuter = VMDCloseableExecuter(closeable: closeable)
             }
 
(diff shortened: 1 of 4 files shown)


frobert11 (approved):

trikot-viewmodels-declarative-flow/swift/swiftui/Components/Snackbar/VMDSnackBarStateObject.swift
```diff
@@ -11,16 +11,16 @@ public class VMDSnackBarStateObject: ObservableObject {
 
     public func startWatching(_ flow: VMDFlow<VMDSnackbarViewData>) {
         flow.watch { [weak self] viewData, closeable in
-            self?.message = viewData.message
```
npresseault, line 14: I think we should do a guard here instead.

npresseault (approved):

trikot-viewmodels-declarative-flow/swift/swiftui/Components/Snackbar/VMDSnackBarStateObject.swift
laurentboileau: If we do a guard and early exit, the closeableExecuter may not be set and we may never stop subscribing to changes.

npresseault: How about we guard but still unsubscribe?
Receiving null should never happen as on kotlin side the generic is typed non-null but it'd be better to just skipping the event than showing an empty snackbar.

laurentboileau: We will:

- Assign the closeableExecutor value as soon as possible
- After that assignment, early exit if the view data is nil

PR #185, opened 2023-09-20
VMDViewModel Flow: Fix Swift isHidden binding

hugolefrancois:
## Description

I found a regression in VMDViewModel Flow SwiftUI components, the field "isHidden" is no longer binded.

I choose to pass the 2 fields to the vmdModifier instead of observing the VM in the ViewModifier to avoid any small performance issue of observing the propertyWillChange 2 times for each component.

## How Has This Been Tested?

Tested in my project.

## Types of changes


npresseault (approved):
A bit more verbose but if that fixes it and keeps good performance 👍

PR #200, opened 2023-10-26
Fix/prevent color name clash between swiftui and trikot.viewmodel font class

marclefrancois:
## Description

In the case where a Trikot based project is currently migrating from trikot.viewmodels to trikot.viewmodel-declarative there is a name clash in VMDHtmlText.swift between SwiftUI.Font and TRIKOT_FRAMEWORK_NAME.Font

## Motivation and Context

To fix a blocking problem for projects that use both versions of the viewModel

## How Has This Been Tested?

Ran the sample app and tested in my projects the changes

## Types of changes


laurentboileau (approved):
Very nice. Thank you for fixing. We encountered the same issue in our project.

PR #202, opened 2023-11-09
Compose android tv bindings

js-bonin:
I added a module adding Compose TV bindings and showcasing some of the new components available in Compose TV like the Carousel, NavigationDrawer and TvLazyColumn / TvLazyRow.

## Motivation and Context
The Compose TV dependency is still in alpha but soon to be release so if we start a new project that also supports TV we will have the bindings available.

## How Has This Been Tested?
I added a TV specific activity in the sample to showcase the bindings

## Types of changes
- [ x] New feature (non-breaking change which adds functionality)

## Result
https://github.com/mirego/trikot/assets/9321287/2b8598d1-02db-49d9-8ebb-86f7ed6893e3


npresseault (approved):

trikot-viewmodels-declarative-flow/compose-tv-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/VMDButtonTv.kt
```diff
@@ -0,0 +1,61 @@
...
+import androidx.tv.material3.ExperimentalTvMaterial3Api
+import com.mirego.trikot.viewmodels.declarative.components.VMDButtonViewModel
+import com.mirego.trikot.viewmodels.declarative.compose.extensions.isOverridingAlpha
+import com.mirego.trikot.viewmodels.declarative.compose.extensions.observeAsState
+import com.mirego.trikot.viewmodels.declarative.compose.extensions.vmdModifier
+import com.mirego.trikot.viewmodels.declarative.content.VMDContent
+
+@ExperimentalTvMaterial3Api
+@Composable
+fun <C : VMDContent> VMDButtonTv(
```
mathieularue, line 29: Should we put these views in their own package instead of suffixing them with Tv? We could reference them using only the import (like the material 3 version of the views vs the standard ones)

mathieularue (commented):
👍🏻

PR #206, opened 2023-12-06
Trikot analytics revamp

sdetilly:
## Description
We deprecate the dependecy to trikot-streams with the following changes

- Deprecate `fun distinctAppId(): Promise<String>` and add `suspend fun distinctAppId(): String`
- Deprecate `AnalyticsService.trackEvent` and `AnalyticsService. identifyUser`, which uses publishers and replace them with functions that don't use publishers.
- Deprecate `TrackableViewModelAction`, because it uses Publishers in its constructor in favour of `SimpleTrackableViewModelAction`

## Motivation and Context

Newer projects should use Coroutines and flows, not trikot-streams.

## How Has This Been Tested?

By running the trikot-viewmodels sample

## Types of changes

### DIFF
--- a/trikot-analytics/firebase-ktx/src/main/java/com/mirego/trikot/analytics/FirebaseAnalyticsService.kt
+++ b/trikot-analytics/firebase-ktx/src/main/java/com/mirego/trikot/analytics/FirebaseAnalyticsService.kt
@@ -4,6 +4,9 @@ import android.content.Context
 import android.os.Bundle
 import com.google.firebase.analytics.FirebaseAnalytics
 import com.mirego.trikot.streams.reactive.promise.Promise
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.resume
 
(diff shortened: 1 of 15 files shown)


trikot-analytics/firebase-ktx/src/main/java/com/mirego/trikot/analytics/FirebaseAnalyticsService.kt
```diff
@@ -40,6 +49,23 @@ class FirebaseAnalyticsService(context: Context, analyticsEnabled: Boolean = tru
...
     }
 
+    override suspend fun distinctDeviceId() = suspendCancellableCoroutine { continuation ->
+        firebaseAnalytics.appInstanceId
+            .addOnCompleteListener { task ->
+                val result = task.result
+                if (task.isSuccessful && result != null) {
+                    continuation.resume(result)
+                } else if (!task.isSuccessful) {
+                    continuation.cancel(Throwable(task.exception))
```
npresseault, line 59: Typed exceptions might help catching specific errors here.

also, why not resumeWithException

PR #210, opened 2023-12-15
refactor: Set Compose modifier as first optional parameter

nicolashaan:
## Description

For Jetpack Compose components, update parameters order so that mandatory parameters come before `modifier` parameters. 

## Motivation and Context

According to [Jetpack Compose guidelines](
https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter):
_Element functions MUST accept a parameter of type Modifier. This parameter MUST be named “modifier” and MUST appear as the first optional parameter in the element function's parameter list._ 

Convention seems to put mandatory parameters before `modifier` parameter.  (See official Compose components such as `Text()`, `Image()`)...
Rational: 
This allows to use the component by passing the `viewModel` parameter without specifying parameter name:
```
VMDButton(viewModel) {
...
}
```
before:
```
VMDButton(viewModel = viewModel) {
...
}
```
## How Has This Been Tested?
It compiles. 

## Types of changes

Technicaly a breaking change as it may break existing calls which don't specify parameter names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Upcoming
 
+### Breaking Changes
+
+- [VMD] Change parameters order for some Compose components
+
 ### Updates
 
 - Gradle to `8.4`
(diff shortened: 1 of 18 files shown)


npresseault (commented):
100% agree with the change, but this will break binary compatibility. Did you run apiDump for the binary compatibility validator?

Please add a line in the CHANGELOG as well.

npresseault:
> Technicaly a breaking change as it may break existing calls which don't specify parameter names.
yes, source compatibility is preserved if you use explicit parameter names but binary compatibility won't.

PR #212, opened 2024-01-05
Remove deprecated Foundation.Date in internal modules

MathieuLegault1:
## Description

We already deprecated the use of trikot `Foundation.Date` but some modules were still using it (internally). This PR changes those usage for `kotlinx-datetime`.

## Motivation and Context

Since removing it internally doesn't affect the api that we expose there is no reason not to remove it. It could be useful if we ever decide to hard deprecate it also

## How Has This Been Tested?

This was tested using the already existing unit test

## Types of changes

- [x] Remove deprecated usage


trikot-viewmodels/databinding/api/databinding.api
```diff
@@ -1,11 +1,3 @@
-public abstract interface class androidx/databinding/DataBindingComponent {
```
MathieuLegault1, line 1: I'm not too sure why this is updated. Perhaps someone forgot to run apiDump after some changes ?

npresseault (approved):

PR #213, opened 2024-01-08
Fix possibly huge memory leak caused by long lived CancellableManagerProvider

marclefrancois:
## Description

A long lived CancellableManagerProvider could easily end up with very many “cancelled” Cancellable in it’s internal CancellableManager because this is only cleared when the Provider itself is cancelled.

This can cause huge memory leaks in some cases, for example in the DebounceProcessor

Since there is only ever one “not cancelled” CancellableManager provided by the CancellableManagerProvider, there is actually no need to keep the previously provided, now cancelled, CancellableManager.

We now simply use the internalCancellableManagerRef.

Also took the oportunity to make cancelPreviousAndCreate hopefully atomic be exposing and using getAndSet on AtomicReference

## Motivation and Context

In certain realtime application, usage of `debounce()` on long lived, shared publishers could cause immense memory leaks

## How Has This Been Tested?

- Tested in our application
- Ran the unit tests

## Types of changes

### DIFF
--- a/trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManagerProvider.kt
+++ b/trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManagerProvider.kt
@@ -3,18 +3,14 @@ package com.mirego.trikot.streams.cancellable
 import com.mirego.trikot.foundation.concurrent.AtomicReference
 
 class CancellableManagerProvider : Cancellable {
-    private val cancellableManager = CancellableManager()
     private val internalCancellableManagerRef = AtomicReference(CancellableManager())
 
-    fun cancelPreviousAndCreate(): CancellableManager {
-        internalCancellableManagerRef.value.cancel()
-        return CancellableManager().also {
-            internalCancellableManagerRef.setOrThrow(internalCancellableManagerRef.value, it)
-            cancellableManager.add(it)
+    fun cancelPreviousAndCreate(): CancellableManager =
+        CancellableManager().also {
+            internalCancellableManagerRef.getAndSet(it).cancel()
         }
-    }
 
     override fun cancel() {
-        cancellableManager.cancel()
+        internalCancellableManagerRef.value.cancel()
     }
 }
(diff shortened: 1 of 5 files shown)


npresseault (approved):

smatte (changes requested):
With this change we change the following behavior: if I create a new CancellableManger  and the provider has been cancelled then the returned manager needs to be cancelled.

Boubalou (approved):

trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManagerProvider.kt
```diff
@@ -1,20 +1,29 @@
...
+    private val isCancelled = AtomicReference(false)
 
     fun cancelPreviousAndCreate(): CancellableManager {
-        internalCancellableManagerRef.value.cancel()
-        return CancellableManager().also {
-            internalCancellableManagerRef.setOrThrow(internalCancellableManagerRef.value, it)
-            cancellableManager.add(it)
+        return CancellableManager().also { cancellableManager ->
+            internalCancellableManagerRef.getAndSet(cancellableManager).cancel()
+            serialQueue.dispatch {
```
marclefrancois, line 14: The usage of the serialQueue is to ensure sequential cancellation of currently returned ref

trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManagerProvider.kt
```diff
@@ -1,20 +1,29 @@
...
+                if (isCancelled.value) {
+                    cancellableManager.cancel()
+                }
+            }
         }
     }
 
     override fun cancel() {
-        cancellableManager.cancel()
+        serialQueue.dispatch {
```
marclefrancois, line 23: The usage of the serialQueue is to ensure sequential cancellation of currently returned ref

trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManagerProvider.kt
marclefrancois: This essentially mimicks the CancellableManager internal implementation to be on the safe side

smatte (approved):

PR #221, opened 2024-06-17
FileSystemDataSource ++

hugolefrancois:
## Description

- Add a way to skip read and save data for a specific request. To do so, you have to subclass FileSystemDataSource and override shouldSkipRequest.
- Implement the delete method.

## How Has This Been Tested?

Tested in my project

## Types of changes


trikot-datasources/datasources-flow-filesystem/src/commonMain/kotlin/com/mirego/trikot/datasources/flow/generic/FileSystemDataSource.kt
```diff
@@ -38,10 +44,14 @@ class FileSystemDataSource<R : ExpiringFlowDataSourceRequest, T>(
     }
 
     override suspend fun save(request: R, data: FlowDataSourceExpiringValue<T>?) {
+        if (shouldSkipRequest(request)) {
```
npresseault, line 47: Not sure I understand properly here. What is the difference between throwing and just not saving anything?

npresseault (commented):
Tests?

trikot-datasources/datasources-flow-filesystem/src/commonMain/kotlin/com/mirego/trikot/datasources/flow/generic/FileSystemDataSource.kt
hugolefrancois: Good point, only a bad copy paste. We should do nothing here and not throw.

npresseault (approved):

PR #222, opened 2024-06-21
[Android] Allow the VMDTextStyleProvider to use Composable functions

amartellino:
Allow the VMDTextStyleProvider to call Composable functions. This will allow the TextStyle to use colours that change depending on the app theme.

## Description
Make VMDTextStyleProvider functions Composable so we're able to call Composable functions.

## Motivation and Context
The current implementation only allows a `TextStyle` to use static colours. In the Sports (TSN/RDS) app, we want the styles to use colours that change depending on the app theme (like use a light font colour if the app is in dark theme or vice-versa)

## How Has This Been Tested?
I made the `SampleTextStyleProvider` in the sample app use `LocalContentColor.current` (which is a Composable) when building the `TextStyle` for  `SampleTextStyleResource.HIGHLIGHTED`.

## Types of changes


npresseault (approved):
Just dont forget to update the changelog since this will be a breaking change.

PR #229, opened 2025-01-14
Add cleanCancelledChildren on CancellableManager

CharlesMcD:
## Description

Introduce the possibility to clean `Cancellable`s that have already been cancelled in a `CancellableManager`.

## Motivation and Context

It's possible to have a global `CancellableManager` that accumulates a huge number of child `Cancellable` during the lifecycle of an app. When we cancel the massive list of `Cancellable`, it can be very performance heavy and can freeze on some lower-end devices.

We can now manually clean up the `queueList` of a `CancellableManager` by choosing the right moment to do it.

I also wanted to make sure that it could support Promises. Previously, a finished** Promise was still kept in the `queueList` until the provided `CancellableManager` was cancelled. (I like to call it the parent CancellableManager)
Now, a Promise that has received a result is considered "cancelled" and ready to be cleaned.

> ** a promise is finished after `onNext`, `onError` or `onCompleted`.

For a `Cancellable` to be cleaned in the `queueList`, the new interface 
```
interface VerifiableCancelledState {
    val isCancelled: Boolean
}
```
 must be implemented.

`CancellableManager` and `CancellableManagerProvider` already implement `VerifiableCancelledState`, but the interface is public in order to be implemented with your custom Cancellable classes. 

## How Has This Been Tested?

It has been tested in a client app.

We can see a big difference in execution time by cleaning periodically our big `CancellableManager` ahead of time. 
The difference is particularly noticeable on lower-end devices like old Android TVs.

```
public fun doTest() {
    val parentCancellableManager = CancellableManager()

    val cancellableList = AtomicListReference<CancellableManager>()

    for (i in 1..5000) {
        val cancellableManager = CancellableManager()
        parentCancellableManager.add(cancellableManager)

        cancellableList.add(cancellableManager)
    }

    cancellableList.value.forEach {
        it.cancel()
    }

    // In a real life scenario, cleaning should not be done right before cancelling the parent CancellableManager.
    // It's up to the application to determine the right timing to optimize the performance.
    parentCancellableManager.cleanCancelledChildren()

    val elapsed = measureTime {
        parentCancellableManager.cancel()
    }

    println("Time elapsed: ${elapsed.inWholeMilliseconds} milliseconds")
}
```

## Types of changes

### DIFF
--- a/trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManager.kt
+++ b/trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManager.kt
@@ -1,18 +1,20 @@
 package com.mirego.trikot.streams.cancellable
 
 import com.mirego.trikot.foundation.concurrent.AtomicListReference
-import com.mirego.trikot.foundation.concurrent.AtomicReference
 import com.mirego.trikot.foundation.concurrent.dispatchQueue.SynchronousSerialQueue
+import kotlinx.atomicfu.atomic
 
-class CancellableManager : Cancellable {
+class CancellableManager : Cancellable, VerifiableCancelledState {
     private val serialQueue = SynchronousSerialQueue()
     private val queueList = AtomicListReference<Cancellable>()
-    private val isCancelled = AtomicReference(false)
+
+    private var isCancelledDelegate = atomic(false)
+    override var isCancelled: Boolean by isCancelledDelegate
 
     fun <T : Cancellable> add(cancellable: T): T {
         queueList.add(cancellable)
 
-        if (isCancelled.value) {
+        if (isCancelled) {
             doCancelAll()
         }
         return cancellable
@@ -28,8 +30,22 @@ class CancellableManager : Cancellable {
         )
     }
 
+    fun cleanCancelledChildren() {
+        serialQueue.dispatch {
+            val value = queueList.value
+            val alreadyCancelledList = value.filter {
+                when (it) {
+                    is VerifiableCancelledState -> it.isCancelled
+                    else -> false
+                }
+            }
+
+            queueList.removeAll(alreadyCancelledList)
+        }
+    }
+
     override fun cancel() {
-        if (isCancelled.compareAndSet(isCancelled.value, true)) {
+        if (isCancelledDelegate.compareAndSet(isCancelledDelegate.value, true)) {
             doCancelAll()
         }
     }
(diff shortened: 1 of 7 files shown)


trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/promise/Promise.kt
```diff
@@ -20,50 +22,71 @@ class Promise<T> internal constructor(
...
+
+    /**
+     * When a result is received, we want to make sure we're considered as "cancelled" by the provided cancellableManager if any.
+     * This allows to use `CancellableManager::cleanCancelledChildren` with promises to clean up finished promises
+     * and avoids keeping a reference on the promise, which is finished anyway. No further results will be emitted.
+     *
+     * We don't want the `onParentCancellation::cancel()` to run since the provided cancellableManager was not cancelled,
+     * so we don't call `cancel()` directly, but instead modify the `isCancelled` value manually.
+      */
+    private fun onResultReceived() {
```
npresseault, line 53: Is there any way we can test that in https://github.com/mirego/trikot/blob/master/trikot-streams/streams/src/commonTest/kotlin/com/mirego/trikot/streams/reactive/promise/PromiseCancellationTests.kt ?

trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManager.kt
```diff
@@ -28,8 +30,22 @@ class CancellableManager : Cancellable {
         )
     }
 
+    fun cleanCancelledChildren() {
```
npresseault, line 33: Thinking out loud here. I wonder if that's not something we should do automatically in the `add()`

npresseault (approved):

thermech (approved):
After comments

alarochelle (approved):

trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/cancellable/CancellableManager.kt
alarochelle: Yeah.. I don't see why not, that would keep it leaner and cleaner at all times, making potential manual cleanups faster.

CharlesMcD: After discussion we determined that it might be too heavy to execute automatically in the `add()` if the `queueList` is really big. We can have a lot of `Cancellable` that will never be cleanable in the list, so you'll have to iterate through them every time for nothing.

So ideally solution is to let the owner of the `CancellableManager` call `cleanCancelledChildren()` at the right moment.

trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/promise/Promise.kt
CharlesMcD: `onResultReceived` only affect internal stuff that are not accessible outside of the Promise.

Unfortunately I don't think this part is testable via unit test unless we make some attributes public only for testing purposes, which I don't think is ideal.

PR #232, opened 2025-04-30
Use URLAnnotation for proper accessibility on Android.

AntoineGagnon:
## Description

Using URLAnnotation will allow the VoiceOver to provide "actions" for each of the links in a section of text. 

## How Has This Been Tested?

Tested using the components preview.

## Types of changes
 
### DIFF
--- a/trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/internal/html/HtmlText.kt
+++ b/trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/internal/html/HtmlText.kt
@@ -35,9 +35,11 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.text.ExperimentalTextApi
 import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.TextLayoutResult
 import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.UrlAnnotation
 import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.style.TextDecoration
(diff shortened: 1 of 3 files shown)


trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/internal/html/HtmlText.kt
```diff
@@ -46,9 +48,9 @@ import androidx.compose.ui.unit.sp
 import androidx.compose.ui.unit.takeOrElse
 import androidx.core.text.HtmlCompat
 
-private const val URL_TAG = "url_tag"
 private val LIST_REGEX = "(<li\\s*[^>]*>)(.*)(</li>)".toRegex()
 
+@OptIn(ExperimentalTextApi::class)
```
npresseault, line 53: Since we're in a library, it's a better practice to propagate the optIn instead of opting in here.
So `@ExperimentalTextApi` instead of `@OptIn(ExperimentalTextApi::class)`

trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/internal/html/HtmlText.kt
```diff
@@ -107,6 +109,7 @@ private fun String.addBullets() =
         "$startTag&nbsp;&nbsp;&nbsp;&nbsp;• $inside$endTag"
     }
 
+@OptIn(ExperimentalTextApi::class)
```
npresseault, line 112: Same idea

npresseault (approved):

trikot-viewmodels-declarative-flow/compose-flow/src/main/kotlin/com/mirego/trikot/viewmodels/declarative/compose/viewmodel/internal/html/HtmlText.kt
AntoineGagnon: Good catch, I've updated it ✅

PR #233, opened 2025-05-09
Fix - combine race condition

aMarquisMirego:
## Description

Adds `set` methods to `AtomicReference` and `AbstractAtomicListReference` for direct value updates instead of doing removeAll() then the equivalent of addAll().
This change is to avoid potential race conditions when subscribing to multiple publishers.

```kotlin
publishersResult.removeAll(publishersResult.value)

repeat(publishers.size + 1) {
    publishersResult.add(PublisherResult())
}
``` 
to
```kotlin
val newPublisherResultsList = mutableListOf<PublisherResult<T>>()
repeat(publishers.size + 1) {
    newPublisherResultsList.add(PublisherResult())
}

publishersResult.set(newPublisherResultsList)
``` 

Also fixed a potential bug with .safeCombine() with 2 publishers in parameters. It doesn't seem to change anything but it was definitely wrong to typecast as T instead of R1 and R2.
## Motivation and Context
It should fix a race condition that can happen when we have multiple publishers subscribed.

## How Has This Been Tested?
Since its a race condition, it is pretty hard to test, but a unit test has been added

## Types of changes

### DIFF
--- a/trikot-foundation/trikotFoundation/src/commonMain/kotlin/com/mirego/trikot/foundation/concurrent/AbstractAtomicListReference.kt
+++ b/trikot-foundation/trikotFoundation/src/commonMain/kotlin/com/mirego/trikot/foundation/concurrent/AbstractAtomicListReference.kt
@@ -14,8 +14,4 @@ abstract class AbstractAtomicListReference<T> {
             mutateWhenReady(block)
         }
     }
-
-    fun set(newValue: List<T>) {
-        internalReference.set(newValue)
-    }
 }
--- a/trikot-foundation/trikotFoundation/src/commonMain/kotlin/com/mirego/trikot/foundation/concurrent/AbstractAtomicListReference.kt
+++ b/trikot-foundation/trikotFoundation/src/commonMain/kotlin/com/mirego/trikot/foundation/concurrent/AbstractAtomicListReference.kt
@@ -14,4 +14,8 @@ abstract class AbstractAtomicListReference<T> {
             mutateWhenReady(block)
         }
     }
+
+    fun set(newValue: List<T>) {
+        internalReference.set(newValue)
+    }
 }
--- a/trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/processors/CombineLatestProcessor.kt
+++ b/trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/processors/CombineLatestProcessor.kt
@@ -80,11 +80,12 @@ internal class CombineLatestProcessor<T>(
         private fun subscribeToCombinedPublishersIfNeeded() {
             if (hasSubscribed.compareAndSet(false, true)) {
                 val cancellableManager = cancellableManagerProvider.cancelPreviousAndCreate()
-                publishersResult.removeAll(publishersResult.value)
-
+                val newPublisherResultsList = mutableListOf<PublisherResult<T>>()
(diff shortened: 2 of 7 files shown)


trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/processors/CombineLatestProcessor.kt
```diff
@@ -80,12 +80,13 @@ internal class CombineLatestProcessor<T>(
         private fun subscribeToCombinedPublishersIfNeeded() {
             if (hasSubscribed.compareAndSet(false, true)) {
                 val cancellableManager = cancellableManagerProvider.cancelPreviousAndCreate()
-                publishersResult.removeAll(publishersResult.value)
-
+                val newPublisherResultsList = mutableListOf<PublisherResult<T>>()
```
npresseault, line 83: If we're to never mutate the publishersResult values, why don't we use an AtomicReference (not a List) and assign the new reference?

trikot-foundation/trikotFoundation/src/commonMain/kotlin/com/mirego/trikot/foundation/concurrent/AbstractAtomicListReference.kt
```diff
@@ -14,4 +14,8 @@ abstract class AbstractAtomicListReference<T> {
             mutateWhenReady(block)
         }
     }
+
+    fun set(newValue: List<T>) {
```
npresseault, line 18: This should not be necessary if we don't use AtomicListReference

trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/processors/CombineLatestProcessorExtensions.kt
```diff
@@ -115,8 +115,8 @@ fun <T, R1, R2> Publisher<T>.safeCombine(publisher1: Publisher<R1>, publisher2:
     return (this as Publisher<Any>).combine(listOf(publisher1, publisher2) as List<Publisher<Any>>)
         .filter { list ->
             list[0] as? T != null &&
-                list[1] as? T != null &&
-                list[2] as? T != null
+                list[1] as? R1 != null &&
```
npresseault, line 118: Nice catch, and probably a result of some wild type erasure magic happening here for that to work with the wrong cast.

trikot-streams/streams/src/commonMain/kotlin/com/mirego/trikot/streams/reactive/processors/CombineLatestProcessor.kt
```diff
@@ -80,11 +80,9 @@ internal class CombineLatestProcessor<T>(
         private fun subscribeToCombinedPublishersIfNeeded() {
             if (hasSubscribed.compareAndSet(false, true)) {
                 val cancellableManager = cancellableManagerProvider.cancelPreviousAndCreate()
-                publishersResult.removeAll(publishersResult.value)
+                val newPublisherResultsList = List(publishers.size + 1) { PublisherResult<T>() }
 
-                repeat(publishers.size + 1) {
-                    publishersResult.add(PublisherResult())
-                }
+                publishersResult.set(newPublisherResultsList)
```
npresseault, line 85: use compareAndSet

PR #235, opened 2025-06-27
Update Kotlin to 2.2.0 + other dependencies

mathieularue:
## Description
We update Kotlin to 2.2.0 and kotlinx-datetime to 0.7.0, plus other dependencies like Gradle, KSP, coroutines, etc.

## Motivation and Context
When updating the kotlinx-datetime library to version 0.7.0 in our project, we encountered runtime incompatibilities with the current version used in Trikot. In order to fix the issue, we update Kotlin to 2.2.0 and we use the new Clock and Instant classes from kotlin.time (as suggested [here](https://github.com/Kotlin/kotlinx-datetime?tab=readme-ov-file#deprecation-of-instant)).

## How Has This Been Tested?
In our project using Maven local

## Types of changes

ABI version is not compatible with earlier version of Kotlin, so at least 2.2.0 must be use.


mathieularue:
Waiting for [this](https://github.com/square/okio/commit/24e3c84dc2b39bbf19806960e7014de4c67abeba) commit to be released in order to fix unit tests

PR #250, opened 2026-03-10
[VMD] Remove SwiftUIIntrospect dependency

FrancisPepin:
The goal of this PR is to open a discussion.

Currently, `VMDTextField` on iOS uses `SwiftUIIntrospect` in order to set the keyboard return type on iOS 13 and 14 only. iOS 15+ has introduced `.submitLabel` so there is no need for introspect.

This creates a transitive dependency on a 1.X version of [SwiftUIIntrospect](https://github.com/siteline/swiftui-introspect), which has now moved to the Apple versioning (26.0.0). This makes it impossible to use the latest version of introspect in a project.

Considering most of our projects are already on a minimum iOS 15, should we:
- Remove Introspect and accept that keyboard return type won't work on iOS 13-14 (current PR)
- Update `SwiftUIIntrospect` in Trikot to `26.0.0` and force every individual project to bump this version as well
- Allow a  '>= 1.0' version instead of a '~> 1.0' version and hope future releases don't break this behaviour in Trikot
- Make Trikot 6.1+ have a min iOS 15, which would remove the need for Introspect as well 
 
## Description

- SwiftUIIntrospect dependency is removed
- VMDTextField on iOS 13 and 14 do not have their keyboard return type set, which is a regression. Introspect would have to be used in a project directly with the same lines of code.

## Motivation and Context

Allows projects to update to the latest `SwiftUIIntrospect` version without being bound by Trikot.

## How Has This Been Tested?

Tested locally on the `/sample` project. Was not able to test on a iOS 13-14 device as I don't have a physical one and it's not available in XCode simulators anymore.

## Types of changes

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


trikot-viewmodels-declarative/sample/ios/Podfile.lock
```diff
@@ -55,10 +52,9 @@ SPEC CHECKSUMS:
...
-  SwiftUIIntrospect: 9f15bf51d1a7582e168a683debca38649f3e152e
-  Trikot: 800d5c7937d265a2ee701dfd49c0a0ce4741438c
-  TRIKOT_FRAMEWORK_NAME: d1e1f18e11fcd960a64712c7780ee069da621d57
+  Trikot: ad8e063808dbf7f9c5dd2c3df7b580720ff86a8a
+  TRIKOT_FRAMEWORK_NAME: b230b4eca34629c9ae4764c796828ec39851275d
 
 PODFILE CHECKSUM: 8057f57ba2a2db77f33517422839afd5282c408e
 
-COCOAPODS: 1.16.2
+COCOAPODS: 1.13.0
```
alarochelle: Shouldn't be downgrading Cocoapods version.

etienneasselin (approved):

alarochelle (approved):

hugolefrancois (approved):

alarochelle:
and don't forget to bump the major version of Trikot on master once merged ;) + add a changelog

MacTHEgenius (approved):
🎉

dominiclabbe (approved):

nlangheit (approved):

